PR #1, opened 2019-08-29
Initial commit

BenchR267:
This initial commit copies all files from the following targets from the Swift package manager repository here to be usable from other projects as well:
TSCBasic, TSCUtility, TSCclibc, TSCLibc, TSCTestSupport, TSCTestSupportExecutable, TSCUtility.

The plan is to eventually move ownership completely to this repository and make SwiftPM depend on it. We're not there yet, especially since it would add a lot complexity to the current bootstrapping of the SwiftPM project at the moment.


Package.swift
```diff
@@ -0,0 +1,78 @@
...
+    dependencies: [
+        
+    ],
+    targets: [
+        
+        // MARK: Tools support core targets
+        
+        .target(
+            /** Shim target to import missing C headers in Darwin and Glibc modulemap. */
+            name: "TSCclibc",
```
ddunbar: Can we call this _TSCClibc or some such?

Package.swift
```diff
@@ -0,0 +1,78 @@
...
+    ],
+    dependencies: [
+        
+    ],
+    targets: [
+        
+        // MARK: Tools support core targets
+        
+        .target(
+            /** Shim target to import missing C headers in Darwin and Glibc modulemap. */
```
ddunbar: `/**` comments aren't actually valid here, can we use `//`?

Utilities/bootstrap
```diff
@@ -0,0 +1,99 @@
...
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2019 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See http://swift.org/LICENSE.txt for license information
+# See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+#
+# -------------------------------------------------------------------------
+# This script is used to bootstrap the tools-support-core package from its
```
ddunbar: We should call this `import` not `bootstrap`.

Bootstrap in our world is mostly used to refer to the process by which a tool which can build itself (e.g. a compiler or a build system) is initially built. As in, "pulling yourself up by your boot straps".

Package.swift
BenchR267: I copied the target declaration right from SwiftPM to match the comments there. Why are `/**` comments not valid here?

Utilities/bootstrap
BenchR267: Makes sense. I wanted to follow the scheme from the other swift repositories. But you're right, it's not building the project.

Package.swift
BenchR267: Same as below, I copied the definition from SwiftPM. Are you worried about the visibility due to the TSC prefix although it's kinda useless outside?

Package.swift
ddunbar: `/**` are "doc comments", which need to get attached to a top-level decl like a class, or a function. You can't (to my knowledge) attach them to a parameter in a call. SwiftPM is wrong, but if that's what it does sure lets be consistent.

Package.swift
ddunbar: I just didn't like the naming convention itself, the lowercase clibc. Not important, you can disregard.

Package.swift
BenchR267: TIL! I thought the only difference is that `/**` allows for multiline comments while `//` is one line only.

ddunbar: *`/*`* allows for multiline.

`/**` is analogous to `///` in that they are Doxygen-derived conventions for annotating doc comments

BenchR267: Oh true, thanks for for clarifying!

PR #7, opened 2019-10-11
Setup proper CMake for TSC

aciidb0mb3r:
There are some incremental build issues to figure out but this mostly works.


CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+cmake_minimum_required(VERSION 3.15.1)
+
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
+    execute_process(
+        COMMAND xcrun --sdk macosx --show-sdk-path
+        OUTPUT_VARIABLE CMAKE_OSX_SYSROOT
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+    SET(CMAKE_Swift_FLAGS "${CMAKE_Swift_FLAGS} -sdk ${CMAKE_OSX_SYSROOT}")
+endif()
```
compnerd: Please dont modify `CMAKE_Swift_FLAGS`, its meant to be used by users only.  You can use `append_compile_options` with a generator expression to add the flags if you must.  (L18 has an example of the generator expression).

CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+project(SwiftTSC LANGUAGES C Swift)
+ 
+set(SWIFT_VERSION 5)
```
compnerd: Why have a variable for this?  Why not just set it directly on L15?

aciidb0mb3r: It's also used in L18

CMakeLists.txt
aciidb0mb3r: Hm, the generator expression doesn't work when CMake is trying to check for the Swift compiler. Is there any other way to fix that?

aciidb0mb3r:
@swift-ci smoke test

aciidb0mb3r:
@swift-ci test

aciidb0mb3r:
@swift-ci test

aciidb0mb3r:
@swift-ci test

PR #26, opened 2020-01-09
Refactor lookupExecutablePath to find executables in PATH even if cwd is not defined

hartbit:


hartbit:
@swift-ci please test

PR #43, opened 2020-02-20
Add a memoize function to simplify caching

hartbit:


hartbit:
@swift-ci test

Sources/TSCBasic/misc.swift
```diff
@@ -229,3 +229,14 @@ extension SystemError: CustomStringConvertible {
         }
     }
 }
+
+/// Memoizes a costly computation to a cache variable.
+public func memoize<T>(to cache: inout T?, build: () throws -> T) rethrows -> T {
```
aciidb0mb3r: Isn't this just `lazy`?

hartbit: It’s lazy, but can be used for throwing operations (`getPackageGraph()`, `getBuildParameters()`, etc...) and for operations taking parameters:

```
memoize(to: &table[key]) { in
    // operation using key
}
```

I wish we had throwing properties 😕

PR #45, opened 2020-02-25
build: workaround older CMake limitations

compnerd:
Add a workaround for older CMake which would not emit the import library
into the correct location.


compnerd:
@swift-ci please smoke test

aciidb0mb3r:
@swift-ci test

PR #46, opened 2020-02-25
build: account for `FoundationNetworking`

compnerd:
When building TSC with a development build of Foundation and using
`-D Foundation_DIR=`, FoundationNetworking would not be found as it
would not be in the library search path.  Explicitly mark the dependency
which also enables better tracking.


aciidb0mb3r:
@swift-ci test

PR #47, opened 2020-02-25
build: workaround CMake 3.15 link failures on Windows

compnerd:
The libraries would be linked without the `-l` prefix on Windows which
works when using absolute paths and C/C++ but not Swift as the driver
will try to read the import library locally.


aciidb0mb3r:
@swift-ci test

PR #51, opened 2020-02-26
Basic: handle root checking on Windows more properly

compnerd:
Roots on Windows are more complicated.  The drive name is not considered
a root, but is a root for the Path API as you can not go any further.
It is however a name and not a path, therefore it is not a root path.
Use that alongwith `PathCchIsRootW` to classify the paths on Windows as
being a root or not.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -164,9 +164,21 @@ public struct AbsolutePath: Hashable {
         return isRoot ? self : AbsolutePath(_impl.dirname)
     }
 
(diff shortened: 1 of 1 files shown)


aciidb0mb3r:
@swift-ci smoke test

aciidb0mb3r:
@swift-ci test

tomerd:
@compnerd should we continue to work on this in light of recent discussions?

compnerd:
The current changes required to bootstrap are in #114.  This should is considered obsoleted.

PR #52, opened 2020-02-26
Basic: replace `resolvingSymlinks` with Foundation

compnerd:
```
realpath() expands all symbolic links and resolves references to
/./, /../ and extra '/' characters in the null-terminated string
named by path to produce a canonicalized absolute pathname.
```

This is effectively the standardized path and resolution of symlinks
along it.


aciidb0mb3r:
@swift-ci smoke test

aciidb0mb3r:
@swift-ci test

tomerd:
@compnerd should we continue to work on this in light of recent discussions?

compnerd:
I think that recent direction sets the plans for something much better in the longer term.  I think that in the short term we should figure out how to get unblocked for getting CI up with s-p-m self-hosting.

tomerd:
cc @shahmishal

PR #54, opened 2020-03-01
Path: Handle Windows paths by converting to an internal UNIX-style representation

troughton:
This is an alternative strategy to the one taken in https://github.com/apple/swift-package-manager/pull/2622; the first commit is just reverting that PR, while the second contains the new changes.

The basic idea is that paths are always stored in UNIX form within `AbsolutePath` and `RelativePath`, and that we convert between Windows paths at `init` at `pathString`. In most cases, removing any UNC or device path prefix, converting backslashes to forward slashes, and prefixing with a root `/` is enough to make a Windows path UNIX-like enough for the rest of the `AbsolutePath`/`RelativePath` functionality to work.

I've only really tested with DOS paths at the moment. If it turns out we need to distinguish between DOS paths, UNC paths, and device paths, we could look at adding a top-level `/windows-device/` or `/windows-unc/` directory to the internal path representation as one way to disambiguate.

cc @compnerd, @hartbit

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -12,12 +12,6 @@ import Foundation
 import WinSDK
 #endif
 
-#if os(Windows)
-private typealias PathImpl = UNIXPath
-#else
-private typealias PathImpl = UNIXPath
-#endif
-
 /// Represents an absolute file system path, independently of what (or whether
 /// anything at all) exists at that path in the file system at any given time.
 /// An absolute path always starts with a `/` character, and holds a normalized
@@ -45,7 +39,7 @@ public struct AbsolutePath: Hashable {
     /// and `RelativePath`; particular file systems may have their own
     /// additional requirements.
     static func isValidComponent(_ name: String) -> Bool {
-        return PathImpl.isValidComponent(name)
+        return name != "" && name != "." && name != ".." && !name.contains("/")
     }
 
     /// Private implementation details, shared with the RelativePath struct.
@@ -62,14 +56,15 @@ public struct AbsolutePath: Hashable {
     /// The input string will be normalized if needed, as described in the
     /// documentation for AbsolutePath.
     public init(_ absStr: String) {
-        self.init(PathImpl(normalizingAbsolutePath: absStr))
+        // Normalize the absolute string.
+        self.init(PathImpl(string: normalize(absolute: absStr)))
     }
 
     /// Initializes an AbsolutePath from a string that may be either absolute
     /// or relative; if relative, `basePath` is used as the anchor; if absolute,
     /// it is used as is, and in this case `basePath` is ignored.
     public init(_ str: String, relativeTo basePath: AbsolutePath) {
-        if PathImpl(string: str).isAbsolute {
+        if str.hasPrefix("/") {
             self.init(str)
         } else {
             self.init(basePath, RelativePath(str))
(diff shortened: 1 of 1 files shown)


hartbit:
I'm not sure this will work. The path needs to be preserved to be passed to file system APIs afterwards. I think we have no choice but to fully support Windows paths.

troughton:
Could you elaborate on what you mean by that and what information you think we wouldn’t be preserving? This does at least work for simple cases - I haven’t tried anything involving multiple drives, network paths, or sum links yet, but I don’t see a fundamental reason why this approach wouldn’t also work for those. 

Note that there is a conversion _out_ of the internal form when we generate the pathString - we don’t pass the internal representation directly to system APIs.

hartbit:
Well, for example, it seems you drop the drive letter and other UNC prefixes. If I initialize `AbsolutePath` with a path on a drive that is not the current drive, I doubt this will work.

troughton:
The drive letter is still in the path - I only drop the `\\?` or `\\`, so e.g. `C:\Folder\File.swift` becomes `/C:/Folder/File.swift`. The main issue I could see is that it doesn’t distinguish as-is between local and UNC paths, but that could be resolved resolved by name spacing at the root level (e.g. `/local/rest-of-path` or `/unc/rest-of-path`).

hartbit:
I understand better now, thanks for explaining! Unfortunately, this feels like a bit of hack, and I think it might be better to truly support Windows paths. I'm working on that, leveraging Windows APIs to avoid reinventing the wheel.

aciidb0mb3r:
I am also thinking that this approach might end up making it harder to do the right thing. Given the fact that @hartbit is already working on properly handling paths using Windows APIs, it might not be worth handling all the complexity we'll be introducing with this approach.

troughton:
My view is that this is a very simple approach and shouldn’t be difficult to eventually replace - most of the complexity in the diff is just the first commit reverting @hartbit’s refactoring, and the second commit mostly amounts to a few lines of string manipulation.

This does have a few advantages over a separate implementation, however: it’s much easier to maintain, can largely share the same test suite, and will have the same performance characteristics as on other platforms (plus a small constant factor on certain operations). I’m not convinced it’s worth all the extra work of a Windows-specific implementation when we don’t yet have concrete use cases that this approach doesn’t work for.

With that all said, if @hartbit thinks a “proper” approach wouldn’t take long to implement and would be performant then I’m totally okay with waiting for that.

tomerd:
@compnerd could you please help review this since you worked on this lately

compnerd:
Processing the same test suite is an interesting idea.  Storing the paths as arcs and then converting to the FSR is a very reasonable thing to do, and really effectively what Foundation does to ensure that different path separators are not a concern.  I think that rebasing this and getting it working to the same state as the current tree would be interesting.

abertelrud:
I agree with @compnerd, I think it's very interesting to consider whether we need the internal representation to be exactly what the OS wants, or something more suitable for path operations that can then (quickly) be transformed into what the OS wants at the time of file system access.  Even with the current implementation of having a string with Unix-style path separators, there can be encoding differences when calling file system functions.  It's just that with Swift String being UTF-8 backed now, it turns out to be a no-op transformation.

It's worth considering that if AbsolutePath and RelativePath contains the OS native representation, then that does get exposed to clients through methods such as `pathString`, etc.  That may be what we want, but it will mean that, for example, `swift package describe` and `swift package dump-package` will yield different output on different platforms.

At some point we do need to deal with paths that have forward slashes on all host OSs, since we need to support at least relative paths that are read from files such as the package manifest (e.g. subpaths of target source directories, of resource files, of module maps, etc).  So I think the decision is whether to do the transformation as each path is constructed, or to do it on file system access.  I don't have a strong opinion yet, but that seems like the main choice, and either way we go has ramifications.

troughton:
For what it's worth, it might be a few weeks before I can get back to any implementation work on this. 

In response to what @abertelrud has raised: my personal opinion is that any relative paths should stay Unix-style for consistency cross-platform, but any absolute paths should always be exposed to the user in a format that they can copy and paste into an Explorer/Finder window. That doesn't really match up with this PR, which would naively expose any absolute paths as the `/`-prefixed internal implementation; however, that's also easy enough to work around.

tomerd:
Hi @troughton, the Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please rebase on top of the `main` and we can reopen.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

PR #56, opened 2020-03-01
Add support for replacing files atomically

keith:
Previously if you tried to write a file to a destination atomically it
would error because moveItem doesn't support overwriting.

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -380,17 +380,9 @@ private class LocalFileSystem: FileSystem {
         if !atomically {
             return try writeFileContents(path, bytes: bytes)
         }
-        try withTemporaryFile(dir: path.parentDirectory, deleteOnClose: false) { temp in
-            do {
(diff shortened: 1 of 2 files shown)


Sources/TSCBasic/FileSystem.swift
```diff
@@ -383,7 +383,11 @@ private class LocalFileSystem: FileSystem {
         try withTemporaryFile(dir: path.parentDirectory, deleteOnClose: false) { temp in
             do {
                 try writeFileContents(temp.path, bytes: bytes)
-                try FileManager.default.moveItem(atPath: temp.path.pathString, toPath: path.pathString)
```
aciidb0mb3r: Maybe we should just switch to Foundation for atomic writes. It can be simplified to:

```swift
try bytes.withData {
    try $0.write(to: URL(fileURLWithPath: path.pathString), options: .atomic)
}
```

keith: Updated!

aciidb0mb3r:
@swift-ci test

PR #58, opened 2020-03-02
Update StringConversions for Windows shell escaping

troughton:
We don't need to escape backslashes on Windows, and multiple double-quotes seem to be standard (https://stackoverflow.com/a/15262019) for escaping other characters. Note that single-quote escaping does not work on Windows shells.


MaxDesiatov:
@swift-ci please test

tomerd:
@compnerd could you please help review this since you worked on this lately

PR #60, opened 2020-03-07
Add Amazon Linux platform detection

drexin:


drexin:
@aciidb0mb3r Unfortunately Amazon Linux does not have an identifiable file like the other fedora flavors. This is the only way I could find to identify it.

drexin:
@swift-ci smoke test

drexin:
@swift-ci test

drexin:
@swift-ci test

drexin:
@swift-ci test

drexin:
@aciidb0mb3r Could you please take a look when you have some time? This is the last thing blocking us from building on Amazon Linux 2.

Sources/TSCUtility/Platform.swift
```diff
@@ -41,7 +41,8 @@ public enum Platform {
             }
             if localFileSystem.isFile(AbsolutePath("/etc/redhat-release")) ||
                localFileSystem.isFile(AbsolutePath("/etc/centos-release")) ||
-               localFileSystem.isFile(AbsolutePath("/etc/fedora-release")) {
+               localFileSystem.isFile(AbsolutePath("/etc/fedora-release")) ||
+               Platform.isAmazonLinux2 {
```
aciidb0mb3r: Given the increasing complexity here, do you mind adding a unit test? I suggest making findCurrentPlatform public and renaming it to _findCurrentPlatform. Then you can add a parameter `fs: FileSystem` with default value of localFileSystem. This should allow you to use InMemoryFileSystem in unit tests.

Sources/TSCUtility/Platform.swift
```diff
@@ -50,6 +51,18 @@ public enum Platform {
         return nil
     }
 
+    private static var isAmazonLinux2: Bool {
```
aciidb0mb3r: I suggest making this a function as it performs IO

Sources/TSCUtility/Platform.swift
```diff
@@ -50,6 +51,18 @@ public enum Platform {
...
+        do {
+            guard let release = try localFileSystem.readFileContents(AbsolutePath("/etc/system-release")).validDescription else {
```
aciidb0mb3r:
```suggestion
            guard let release = try localFileSystem.readFileContents(AbsolutePath("/etc/system-release")).cString else {
```

Sources/TSCUtility/Platform.swift
```diff
@@ -50,6 +51,18 @@ public enum Platform {
...
+                return false
+            }
+
+            return release.starts(with: "Amazon Linux release 2")
```
aciidb0mb3r:
```suggestion
            return release.hasPrefix("Amazon Linux release 2")
```

drexin:
@swift-ci test

drexin:
@swift-ci test

drexin:
@swift-ci test

drexin:
@swift-ci test

drexin:
@aciidb0mb3r Thanks for the review. I addressed your comments, hope the test is what you had in mind.

Tests/TSCUtilityTests/PlatformTests.swift
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+import XCTest
+import TSCBasic
+import TSCTestSupport
+
+import TSCUtility
+
+final class PlatformTests: XCTestCase {
+#if os(Linux)
```
aciidb0mb3r: These tests can be run on all platforms so I suggest dropping this `#if`.

Tests/TSCUtilityTests/PlatformTests.swift
```diff
@@ -0,0 +1,53 @@
...
+        fs = InMemoryFileSystem(files: ["/etc/centos-release": "xxx"])
+        XCTAssertEqual(Platform.linux(.fedora), Platform._findCurrentPlatform(fs))
+
+        fs = InMemoryFileSystem(files: ["/etc/system-release": "Amazon Linux release 2 (Karoo)"])
+        XCTAssertEqual(Platform.linux(.fedora), Platform._findCurrentPlatform(fs))
+    }
+#elseif os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+    
+    func testCurrentPlatformDarwin() {
+        XCTAssertEqual(Platform.darwin, Platform.currentPlatform)
```
aciidb0mb3r: I am not sure this provides a lot of value, I think we can drop it.

aciidb0mb3r:
The tests look great, that's exactly what I had in mind. I just left two minor comments.

Tests/TSCUtilityTests/PlatformTests.swift
drexin: The check executes `uname` to figure out the platform, so when executed on a mac, it will always return `.darwin`.

Tests/TSCUtilityTests/PlatformTests.swift
drexin: See comment above, also I think it‘s valuable to ensure we don‘t regress here.

Tests/TSCUtilityTests/PlatformTests.swift
aciidb0mb3r: Ah. Can you move the switch case into a different function so it is unit testable?

drexin: Ok, will do.

drexin:
@swift-ci test

drexin:
@aciidb0mb3r Updated the PR. So with the new structure I had to remove the darwin test. I could change it further to re-enable it, but that would mean either having a function that statically just return `.darwin`, which doesn't make sense to test, or passing in the uname from outside. Not sure either case is worth testing, but your call 🙂

drexin:
@swift-ci test

aciidb0mb3r:
LGTM, thanks for working on this! Can you also make a PR on the main SwiftPM repo?

Conan-Kudo:
This platform detection is screwy, and I'm definitely certain that Amazon should probably try to fix this.

@mswilson, could you see if Amazon Linux could gain `ID_LIKE="rhel fedora"` in its `os-release(5)` file?

PR #67, opened 2020-04-13
[TSCUtility] Introduce PolymorphicCodable property wrapper

aciidb0mb3r:
This allows encoding and decoding polymorphic types without writing
a bunch of boilerplate code.

### DIFF
--- a/Sources/TSCUtility/PolymorphicCodable.swift
+++ b/Sources/TSCUtility/PolymorphicCodable.swift
@@ -1,0 +1,56 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+/// Allows encoding and decoding known polymorphic types.
+public protocol PolymorphicCodableProtocol: Codable {
+    static var implementations: [PolymorphicCodableProtocol.Type] { get }
+}
+
+@propertyWrapper
+public struct PolymorphicCodable<T: PolymorphicCodableProtocol>: Codable {
+    public let value: T
+
+    public init(wrappedValue value: T) {
+        self.value = value
+    }
+
+    public var wrappedValue: T {
+        return value
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        try container.encode(String(reflecting: type(of: value)))
+        try container.encode(value)
+    }
+
+    public init(from decoder: Decoder) throws {
+        var container = try decoder.unkeyedContainer()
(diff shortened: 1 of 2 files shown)


Sources/TSCUtility/PolymorphicCodable.swift
```diff
@@ -0,0 +1,56 @@
...
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        try container.encode(String(reflecting: type(of: value)))
+        try container.encode(value)
+    }
+
+    public init(from decoder: Decoder) throws {
+        var container = try decoder.unkeyedContainer()
+        let typeCode = try container.decode(String.self)
+        guard let klass = T.implementations.first(where: { String(reflecting: $0) == typeCode }) else {
```
ddunbar, line 37: Should this error or at least assert if there are duplications, to detect programmer error?

Sources/TSCUtility/PolymorphicCodable.swift
```diff
@@ -0,0 +1,56 @@
...
+ Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+/// Allows encoding and decoding known polymorphic types.
+public protocol PolymorphicCodableProtocol: Codable {
+    static var implementations: [PolymorphicCodableProtocol.Type] { get }
```
ktoso, line 13: Could be a Set<> I suppose 🤔

Sources/TSCUtility/PolymorphicCodable.swift
ktoso: if implementations was a Set then we (eventually) could `typeHint` -> `_typeByName` to then `T.allowedImplementations.contains(thatTypeByName)` to avoid that scanning through. The T we need anyway in order to get the `Any.Type as Codable` to call decoding on it

ktoso: I'm exploring another direction for registering types on Joe's hints... I'll report back once I have something.

More based around "intent" or "permissions" i.e. "this type is intended for crossing network" vs. "this one isnt" etc.

Sources/TSCUtility/PolymorphicCodable.swift
aciidb0mb3r: Can't use Set:

Type 'PolymorphicCodableProtocol.Type' does not conform to protocol 'Hashable' and Cannot extend a metatype 'PolymorphicCodableProtocol.Type'

vlm: Since `String(reflecting: )` is needed every time `implementations` is accessed, it might make sense to do a struct representing the `(String, Type)`, and implement a `Hashable` over it. Then it is going to be `Set`-compatible. I am sure you've thought of that, just 2c.

aciidb0mb3r:
@swift-ci test

PR #71, opened 2020-05-14
[FileSystem] Throw "unsupported" rather than fatalError()'ing

DougGregor:
Clients should be able to cope with unsupported behavior in
FileSystems. Give them the chance to.


DougGregor:
@swift-ci please test

PR #69, opened 2020-04-20
Add Process support for workingDirectory

sergiocampama:

PR #74, opened 2020-05-30
Add support for posix_spawn_file_actions_addchdir_np on available Linux platforms

sergiocampama:
This adds support for workingDirectory APIs for TSCBasic.Process on Linux if the system where the library is compiled with has libc6 >= 2.29. Uses the `__GLIBC_PREREQ(2, 29)` check as per @jakepetroules 's suggestion on #69 

Because there are no compile time checks for this version on Swift, I had to add runtime support for the failure mode where the `posix_spawn_file_actions_addchdir_np` function is not available.

### DIFF
--- a/Sources/TSCclibc/include/process.h
+++ b/Sources/TSCclibc/include/process.h
@@ -1,0 +1,12 @@
+#if defined(__linux__)
+
+#include <spawn.h>
+#include <stdbool.h>
+
+// Wrapper method for posix_spawn_file_actions_addchdir_np that fails on Linux versions that do not have this method available.
+int SPM_posix_spawn_file_actions_addchdir_np(posix_spawn_file_actions_t *restrict file_actions, const char *restrict path);
+
+// Runtime check for the availability of posix_spawn_file_actions_addchdir_np. Returns 0 if the method is available, -1 if not.
+bool SPM_posix_spawn_file_actions_addchdir_np_supported();
+
+#endif
--- a/Sources/TSCclibc/process.c
+++ b/Sources/TSCclibc/process.c
@@ -1,0 +1,23 @@
+#if defined(__linux__)
+
+#include <errno.h>
+
+#include "process.h"
+
+int SPM_posix_spawn_file_actions_addchdir_np(posix_spawn_file_actions_t *restrict file_actions, const char *restrict path) {
+#if __GLIBC_PREREQ(2, 29)
+    return posix_spawn_file_actions_addchdir_np(file_actions, path);
+#else
+    return ENOSYS;
+#endif
+}
+
+bool SPM_posix_spawn_file_actions_addchdir_np_supported() {
+#if __GLIBC_PREREQ(2, 29)
+    return true;
+#else
+    return false;
+#endif
+}
+
+#endif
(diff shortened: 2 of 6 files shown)


sergiocampama:
@swift-ci test

sergiocampama:
Tested this using Docker on Ubuntu 20.04 where 2.31 is the default and on Ubuntu 18.04 where the max available version is 2.27 (which doesn't have this API)

Sources/TSCclibc/process.c
```diff
@@ -0,0 +1,21 @@
+#if defined(__linux__)
+
+#include "process.h"
+
+int SPM_posix_spawn_file_actions_addchdir_np(posix_spawn_file_actions_t *restrict file_actions, const char *restrict path) {
+#if __GLIBC_PREREQ(2, 29)
+    return posix_spawn_file_actions_addchdir_np(file_actions, path);
+#else
+    return -1;
```
jakepetroules, line 9: How about `ENOSYS` from errno.h?

> “Function not implemented.” This indicates that the function called is not implemented at all, either in the C library itself or in the operating system. When you get this error, you can be sure that this particular function will always fail with ENOSYS unless you install a new version of the C library or the operating system.

Sources/TSCclibc/include/process.h
```diff
@@ -0,0 +1,11 @@
+#if defined(__linux__)
+
+#include <spawn.h>
+
+// Wrapper method for posix_spawn_file_actions_addchdir_np that fails on Linux versions that do not have this method available.
+int SPM_posix_spawn_file_actions_addchdir_np(posix_spawn_file_actions_t *restrict file_actions, const char *restrict path);
+
+// Runtime check for the availability of posix_spawn_file_actions_addchdir_np. Returns 0 if the method is available, -1 if not.
+int SPM_posix_spawn_file_actions_addchdir_np_supported();
```
jakepetroules, line 9: The name ending in _supported seems to imply this should return a `bool`, otherwise I'd expect something more like `SPM_posix_spawn_file_actions_addchdir_np_support_status` which could return 0, ENOSYS, etc.

Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -211,32 +211,40 @@ class ProcessTests: XCTestCase {
...
+        }
+      #endif
 
-                do {
-                    let process = Process(arguments: ["cat", "file"], workingDirectory: tempDirPath)
-                    try process.launch()
-                    let result = try process.waitUntilExit()
-                    XCTAssertEqual(try result.utf8Output(), "parent")
-                }
+        try! withTemporaryDirectory(removeTreeOnDeinit: true) { tempDirPath in
```
jakepetroules, line 227: `try!` here can be `try`, so that the test case fails if an error is thrown instead of crashing the test runner

Sources/TSCclibc/process.c
sergiocampama: yeah, that looks better

Sources/TSCclibc/include/process.h
sergiocampama: changed to bool instead

Tests/TSCBasicTests/ProcessTests.swift
sergiocampama: don't know how I missed that :p

sergiocampama:
@swift-ci test

sergiocampama:
@swift-ci test

PR #75, opened 2020-06-01
Add support for redirecting stderr to stdout.

sergiocampama:
This is done by expanding the OutputRedirection enum to support a boolean. 

This does mean that the stream enum case will have a required closure that won't be used in the case redirectStderr is set, but shouldn't be much of an eye sore. Also adds some source compatibility methods to avoid breaking current users of the API.

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -129,10 +129,20 @@ public final class Process: ObjectIdentifierProtocol {
     public enum OutputRedirection {
         /// Do not redirect the output
         case none
-        /// Collect stdout and stderr output and provide it back via ProcessResult object
-        case collect
-        /// Stream stdout and stderr via the corresponding closures
-        case stream(stdout: OutputClosure, stderr: OutputClosure)
+        /// Collect stdout and stderr output and provide it back via ProcessResult object. If redirectStderr is true,
+        /// stderr be redirected to stdout.
+        case collect(redirectStderr: Bool)
+        /// Stream stdout and stderr via the corresponding closures. If redirectStderr is true, stderr be redirected to
+        /// stdout.
+        case stream(stdout: OutputClosure, stderr: OutputClosure, redirectStderr: Bool)
+
+        /// Default collect OutputRedirection that defaults to not redirect stderr. Provided for API compatibility.
+        public static let collect: OutputRedirection = .collect(redirectStderr: false)
+
+        /// Default stream OutputRedirection that defaults to not redirect stderr. Provided for API compatibility.
+        public static func stream(stdout: @escaping OutputClosure, stderr: @escaping OutputClosure) -> Self {
+            return .stream(stdout: stdout, stderr: stderr, redirectStderr: false)
+        }
 
         public var redirectsOutput: Bool {
             switch self {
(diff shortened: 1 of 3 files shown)


sergiocampama:
@swift-ci test

Sources/TSCBasic/Process.swift
```diff
@@ -436,16 +456,27 @@ public final class Process: ObjectIdentifierProtocol {
...
-                posix_spawn_file_actions_addclose(&fileActions, pipe[1])
+            posix_spawn_file_actions_addclose(&fileActions, outputPipe[0])
+            posix_spawn_file_actions_addclose(&fileActions, outputPipe[1])
+
+            if outputRedirection.redirectStderr {
+                // If merged was requested, send stderr to stdout.
+                posix_spawn_file_actions_adddup2(&fileActions, 1, 2)
+            } else {
+                // If no redirect was requested, open the pipe for stderr.
+                try open(pipe: &stderrPipe)
```
vlm, line 474: This would leak the descriptors in `outputPipe` if thrown.

Sources/TSCBasic/Process.swift
```diff
@@ -129,10 +129,19 @@ public final class Process: ObjectIdentifierProtocol {
...
         case none
-        /// Collect stdout and stderr output and provide it back via ProcessResult object
-        case collect
-        /// Stream stdout and stderr via the corresponding closures
-        case stream(stdout: OutputClosure, stderr: OutputClosure)
+        /// Collect stdout and stderr output and provide it back via ProcessResult object.
+        case collect(redirectStderr: Bool)
+        /// Stream stdout and stderr via the corresponding closures. If redirectStderr is true, stderr be redirected to
+        /// stdout.
+        case stream(stdout: OutputClosure, stderr: OutputClosure, redirectStderr: Bool)
```
vlm, line 136: If we're changing the API, why not take an opportunity to make it symmetric?
```
  case stream(stdout: OutputClosure?, stderr: OutputClosure?)
```

Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -211,6 +211,28 @@ class ProcessTests: XCTestCase {
         }
     }
 
+    func testStdoutStdErrRedirected() throws {
```
vlm, line 214: Are you planning to add the tests for the new functionality?

Sources/TSCBasic/Process.swift
vlm: Disregard.

Sources/TSCBasic/Process.swift
```diff
@@ -436,16 +456,27 @@ public final class Process: ObjectIdentifierProtocol {
         var outputPipe: [Int32] = [0, 0]
         var stderrPipe: [Int32] = [0, 0]
```
vlm, line 457: nit: I'd recommend putting [-1, -1] as the default values.

Tests/TSCBasicTests/ProcessTests.swift
sergiocampama: wdym? this test case contains redirectStderr test cases by checking that stderr output is empty and that stdout contains stderr as well as stdout... what's missing?

Sources/TSCBasic/Process.swift
sergiocampama: that's a problem with the current code as well, no? same with any of the trys in the rest of the code (including the closes)... while I agree that that should be fixed, that particular problem is not introduced by this change, and I would consider it out of scope (since it's unclear which other things might need to be changed)

Sources/TSCBasic/Process.swift
sergiocampama: changed to -1

Sources/TSCBasic/Process.swift
vlm: indeed

Tests/TSCBasicTests/ProcessTests.swift
vlm: I mean tests for `OutputRedirections.stream(_, _, _)`, that it works.

sergiocampama: added those for streaming as well

sergiocampama:
@swift-ci test

sergiocampama:
@swift-ci test

PR #77, opened 2020-06-11
Fix swift-tools-support-core path in Utilities/import

MaxDesiatov:
Currently the directory is named `swift-tools-support-core`, not `TSC`.


aciidb0mb3r:
@swift-ci test

PR #79, opened 2020-06-15
Revert "Improve FileSystemError by adding associated path"

MaxDesiatov:
This reverts #78, which breaks the public API for both SwiftPM and SourceKit-LSP.


aciidb0mb3r:
@swift-ci test

PR #82, opened 2020-06-30
Sink some support code from llbuild2

dmbryson:


dmbryson:
@swift-ci please test

PR #92, opened 2020-07-19
Allow specifying environment for Process on Windows

egorzhdan:
This is going to be useful for getting the package structure from `Package.swift`:
Currently SwiftPM compiles the package manifest with an `-rpath` linker argument which is not supported on Windows.
On Windows it should instead execute the compiled manifest with an additional `PATH` component representing the `PackageDescription.dll` location.

(corresponding PR with the actual usage in SwiftPM is coming soon)


MaxDesiatov:
@swift-ci please test

PR #96, opened 2020-07-30
Fix Git discovery & execution on Windows

egorzhdan:
This fixes:
* git discovery by searching for `git.exe` instead of `git`
* git invocation by passing an absolute path instead of just an executable name

For an example of usage please see `GitRepositoryProvider.fetch` in the SwiftPM project.

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -340,14 +340,15 @@ public final class Process: ObjectIdentifierProtocol {
         }
 
         // Look for executable.
-        guard Process.findExecutable(arguments[0]) != nil else {
-            throw Process.Error.missingExecutableProgram(program: arguments[0])
+        let executable = arguments[0]
+        guard let executablePath = Process.findExecutable(executable) else {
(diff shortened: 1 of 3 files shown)


Sources/TSCBasic/Process.swift
```diff
@@ -340,14 +340,15 @@ public final class Process: ObjectIdentifierProtocol {
         }
 
         // Look for executable.
-        guard Process.findExecutable(arguments[0]) != nil else {
-            throw Process.Error.missingExecutableProgram(program: arguments[0])
+        let executable = arguments[0]
+        guard let absolutePath = Process.findExecutable(executable) else {
```
aciidb0mb3r, line 344:
```suggestion
        guard let executablePath = Process.findExecutable(executable) else {
```

egorzhdan: Renamed along with the usage

aciidb0mb3r:
@swift-ci smoke test

MaxDesiatov:
@swift-ci please test

PR #98, opened 2020-08-06
Fix build on Windows

egorzhdan:
* SQLite3 was recently added as a dependency, and CMake build scripts needed a couple of fixes
* `dirent` is not available on Windows
* `TSCUtility.dll` was not copied into the installation directory


MaxDesiatov:
@swift-ci please test

PR #102, opened 2020-08-10
build: ignore deprecation warnings on Windows

compnerd:
Adjust the compile flags for TSCBasic on Windows.  Avoid the separate
checks and addition of flags in favour of generator expressions.

PR #103, opened 2020-08-10
TSCBasic: avoid POSIX name on Windows for `strdup`

compnerd:
THis generates a warning, so use the "portable" name `_strdup` instead.
This should be functionally identical.


compnerd:

compnerd:
@swift-ci please test

tomerd:

PR #104, opened 2020-08-10
TSCBasic: improve `TemporaryFile` on Windows

compnerd:
Rather than using the ASCII version of the functions which prevent the
use of unicode characters in the path.  Given that `String` in Swift
internally will deal with the encoding, this allows us to encode to
UTF-16 and use the unicode version of the functions.


compnerd:

compnerd:
@swift-ci please test

tomerd:

PR #105, opened 2020-08-10
TSCBasic: implement support for width in the terminal

compnerd:
This adjusts the path to work on Windows which does not support ncurses
and thus does not have `winsize`.  Additionally, `COLUMNS` is not a
variable that exists on Windows to indicate the console size.


compnerd:

compnerd:
@swift-ci please test

tomerd:

PR #111, opened 2020-08-12
TSCLibc: fix `mkstemps` on Windows

compnerd:


compnerd:

compnerd:
This is really waiting on #110; once that is merged, this can be rebased and merged.  This is pretty important and prevents the manifest processing in s-p-m on Windows.

compnerd:

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

tomerd:
@abertelrud plz help review this one

Sources/TSCLibc/libc.swift
```diff
@@ -75,24 +75,35 @@ public func mkstemps(
...
+  // on other architectures) when the memory is accessed.  This roughly emulates
+  // that by terminating in the case even though it is possible for us to return
+  // an error.
+  guard let template = template else { fatalError() }
+
+  let length: Int = strlen(template)
+
+  // Validate the precondition: the template must terminate with 6 `X` which
+  // will be filled in to generate a unique directory.
+  guard length >= 6, memcmp(template + length - Int(suffixlen) - 6, "XXXXXX", 6) == 0 else {
```
abertelrud, line 90: Looks like this would crash if `suffixlen` is too large.  Or perhaps there's a guard that I'm just not seeing?

Sources/TSCLibc/libc.swift
```diff
@@ -75,24 +75,35 @@ public func mkstemps(
...
+    _set_errno(EINVAL)
+    return -1
+  }
+
+  // Attempt to create file
+  var retries: Int = 100
+  repeat {
+    __randname(template + length - Int(suffixlen) - 6)
```
abertelrud, line 98: Looks like this would crash if `suffixlen` is too large.  Or perhaps there's a guard that I'm just not seeing?

Sources/TSCLibc/libc.swift
compnerd: If `suffixlen` is enough to wrap beyond the limits of `Int`, then the string buffer would have to be beyond 4GiB (though in this case beyond the addressing limits actually).  These functions are trying to emulate the C behavior.  If `suffixlen` is greater than `length` we would actually be accessing invalid memory.

abertelrud: Right, I was talking about the case in which `suffixlen` is greater than the string length.  Since this is emulating C behavior, which all its caveats, it seems reasonable to have the same unsafety.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

Sources/TSCLibc/libc.swift
compnerd: Yeah, even tries to emulate the failure on a nullptr input :).  Hopefully we will soon be able to replace some of these unsafe functions with better alternatives.

compnerd:
@swift-ci please smoke test

Sources/TSCLibc/libc.swift
abertelrud: Right, I saw that.  Crash for crash compatible.  :)  Thanks for your work on this!  Please go ahead and merge after the tests complete.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please test

PR #112, opened 2020-08-12
TSCTestSupport: remove `PseudoTerminal` on Windows temporarily

compnerd:
While swift-package-manager is being brought up on Windows, we do not
have the tests running.  Windows does not have the `pty` concept nor the
supporting APIs.  For now disable the class to allow
swift-tools-support-core to build on Windows with swift-package-manager
which is required to bootstrap swift-package-manager.

### DIFF
--- a/Sources/TSCTestSupport/PseudoTerminal.swift
+++ b/Sources/TSCTestSupport/PseudoTerminal.swift
@@ -11,6 +11,8 @@
 import TSCBasic
 import TSCLibc
 
+#if os(Windows)
+#else
 public final class PseudoTerminal {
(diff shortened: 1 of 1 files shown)


compnerd:

compnerd:
@swift-ci please test

tomerd:
I am fine with this as a temp measure. should we document / spit out a warning or somehow remind ourselves to fix it?

compnerd:
Doesn't hurt: https://bugs.swift.org/browse/SR-13394  This will just pop out anyways once we get further along and try to get the test suite working :)

PR #113, opened 2020-08-15
Fix and use posix_spawn_file_actions_addchdir_np() check for Android

buttaface:
This function and `__GLIBC_PREREQ()` are not available in Bionic.


MaxDesiatov:
@swift-ci please test

tomerd:
cc @compnerd

buttaface:
Thanks for the quick review.

PR #114, opened 2020-08-20
various hacks for paths on Windows

compnerd:
This is sufficient for bootstrapping s-p-m on Windows, although the long
term maintainable solution is to migrate towards Swift System.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -429,8 +429,15 @@ private struct UNIXPath: Path {
 
     var dirname: String {
 #if os(Windows)
-        let dir = string.deletingLastPathComponent
-        return dir == "" ? "." : dir
+        let fsr: UnsafePointer<Int8> = string.fileSystemRepresentation
+        defer { fsr.deallocate() }
+
+        let path: String = String(cString: fsr)
+        return path.withCString(encodedAs: UTF16.self) {
+            let data = UnsafeMutablePointer(mutating: $0)
+            PathCchRemoveFileSpec(data, path.count)
+            return String(decodingCString: data, as: UTF16.self)
+        }
 #else
         // FIXME: This method seems too complicated; it should be simplified,
         //        if possible, and certainly optimized (using UTF8View).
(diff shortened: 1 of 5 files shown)


compnerd:
CC: @egorzhdan

Sources/TSCBasic/Path.swift
```diff
@@ -821,7 +888,15 @@ extension AbsolutePath {
     /// This method is strictly syntactic and does not access the file system
     /// in any way.
     public func contains(_ other: AbsolutePath) -> Bool {
+#if os(Windows)
+        return self._impl.string.withCString(encodedAs: UTF16.self) { lhs in
+            other._impl.string.withCString(encodedAs: UTF16.self) { rhs in
+                PathIsSameRootW(lhs, rhs)
```
egorzhdan, line 894: Perhaps I'm missing something, but this doesn't seem to be equivalent to the non-Windows branch... It doesn't ensure that `other` is a subpath of `self`, right?

compnerd: Yeah, this isn't correct :/ however, seems to allow getting through a bootstrap of s-p-m (which is what this change is really about).  I don't have a problem with replacing it with a different implementation, the goal here is to just get to the point where we can get some form of builds of s-p-m.

compnerd: Turns out this change is not required, lets just drop it.

compnerd:
@swift-ci please test

egorzhdan (approved):

compnerd:
@swift-ci please test

Sources/TSCBasic/Path.swift
```diff
@@ -475,13 +489,17 @@ private struct UNIXPath: Path {
...
+#endif
     }
 
     // FIXME: We should investigate if it would be more efficient to instead
     // return a path component iterator that does all its work lazily, moving
     // from one path separator to the next on-demand.
     //
     var components: [String] {
+#if os(Windows)
+        return string.components(separatedBy: "\\").filter { !$0.isEmpty }
```
compnerd, line 501: Technically, this should use `PathSkipRootW` and then split on `\` and `/`.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
I still dislike this,, but this should hopefully unblock the work necessary to support Windows.  Discussed offline with @abertelrud and @neonichu about using this as a crutch for the time being to enable us to more rapidly support Windows.

PR #116, opened 2020-08-20
build: synchronise CMake and Package.swift for Windows

compnerd:
Add the additional link dependency on Windows and a macro definition to
avoid some spew from the library build.  Note that this is using an OS
specific block to alter the settings rather than the inline definition
since older versions of s-p-m may crash on the application of the
setting.

This requires bumping the tools version to 5.2.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.2
 
 /*
  This source file is part of the Swift.org open source project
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @tomerd

compnerd:
@swift-ci please test

abertelrud:
Seems that the Linux test is still using 5.1 tools:
`'/home/buildnode/jenkins/workspace/pr-swift-tools-support-core-linux/branch-master/swift-tools-support-core' is using Swift tools version 5.2.0 but the installed version is 5.1.0`

So that's not a problem with this PR but with the testing infrastructure.  Not sure who maintains that and how we get it updated.

buttaface:
I've had to [update to 5.2 tools for Android too](https://github.com/termux/termux-packages/blob/master/packages/swift/llbuild-Package.swift.patch#L7), as that's when support for that platform was added to SPM, so I'd like this too.

abertelrud:
Thanks; I have also filed a bug report on the CI testing issue.

shahmishal:
@swift-ci test

abertelrud:
Thanks for updating it so quickly, @shahmishal !  @compnerd, this can now be merged when you think it's appropriate.

PR #118, opened 2020-08-27
Parse compiler messages properly on Windows

egorzhdan:
Currently SwiftPM dumps raw JSON data while compiling on Windows instead of the actual build status. This happens because `JSONMessageStreamingParser` splits the output lines by `\n` without taking `\r` into account on Windows.
After this change the compiler messages are parsed and displayed properly.

### DIFF
--- a/Sources/TSCUtility/JSONMessageStreamingParser.swift
+++ b/Sources/TSCUtility/JSONMessageStreamingParser.swift
@@ -101,6 +101,11 @@ private extension JSONMessageStreamingParser {
 
     /// Throwing implementation of the parse function.
     func parseImpl<C>(bytes: C) throws where C: Collection, C.Element == UInt8 {
+#if os(Windows)
+        let carriageReturn = UInt8(ascii: "\r")
+        let bytes = bytes.filter { $0 != carriageReturn }
+#endif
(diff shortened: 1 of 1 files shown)


egorzhdan:
@swift-ci please test

egorzhdan:
@swift-ci please test

egorzhdan:
cc @compnerd

compnerd:
Thanks for identifying the issue!

What happens if it wants a raw carriage return?  I suppose it will be escaped in the JSON right?

egorzhdan:
Yeap, it gets escaped the same way as `\n`.

compnerd:
Hmm, I wonder if there is a way to sink this lower so that when the processing is done, it splits on `\r\n` and `\n`.

compnerd:
@swift-ci please test

egorzhdan:
It's certainly possible, that would just be a bit more complex. This fix is simple and just works.
I can move the `\r` handling logic to the parsing stage though, if you think this is hacky.

compnerd:
It definitely would be interesting to know how more complicated it is.  If its not very complicated, that definitely would be nicer I think.

PR #120, opened 2020-09-02
Move LLVM Bitstream Utils from swift-driver and add a Swift/Clang Serialized Diags Reader

owenv:
The first commit in this PR copies over `Bits.swift` and `Bitstream.swift` unmodified from swift-driver. These are general utilities for reading files using the LLVM bitstream container format. The second commit is new code to support reading serialized diagnostics (.dia) files emitted by Swift and Clang. The intent is that this can be used to improve how SPM handles diagnostics emitted by the compiler during a build.

If this is merged, I'll follow up to remove the copied files in swift-driver and use the TSC versions instead.


owenv:
@swift-ci please test

aciidb0mb3r:
👍

neonichu (approved):

PR #122, opened 2020-09-05
Add support for Process stdin stream

sergiocampama:


sergiocampama:
@swift-ci please test

Sources/TSCBasic/Process.swift
```diff
@@ -113,6 +113,9 @@ public struct ProcessResult: CustomStringConvertible {
     }
 }
 
+/// Alias to make the type's name closer to its purpose.
+public typealias ProcessStdinStream = OutputByteStream
```
abertelrud, line 117: I'm not sure whether the alias makes things more clear or less clear here.  I understand the intent, but for clients who are already familiar with `OutputByteStream`, the introduction of what seems like a unique type is a bit confusing as well.  I'm not strictly opposed, just mildly concerned about the understandability.  I don't think that calling it an output byte stream is wrong, since it's an output from the perspective of the parent process.

Maybe in a future API overhaul we should actually call them `ReadableByteStream` and `WritableByteStream` or something like that.

Sources/TSCBasic/Process.swift
```diff
@@ -531,6 +547,8 @@ public final class Process: ObjectIdentifierProtocol {
             }
         }
       #endif // POSIX implementation
+
+        return stdinStream
```
abertelrud, line 551: Are we counting on `deinit` closing the stream for the existing clients (or any new clients that ignore the return value)?  Is there any risk in making the existing function return a value that has to be released, for example by turning what would have otherwise been a memory leak (e.g. a captured output value in a block or anything like that) into something that's now holding open the input possibly indefinitely (which depending on the subprocess could mean that it doesn't actually complete)?

Sources/TSCBasic/Process.swift
```diff
@@ -731,11 +749,15 @@ private func open(pipe: inout [Int32]) throws {
 }
 
 /// Close the given fd.
-private func close(fd: inout Int32) throws {
-    let rv = TSCLibc.close(fd)
-    guard rv == 0 else {
-        throw SystemError.close(rv)
+private func close(fd: Int32) throws {
+    func innerClose(_ fd: inout Int32) throws {
```
abertelrud, line 753: Is this just cleanup to avoid having to pass the file descriptor by address, or is there another reason for changing this function?

sergiocampama: yeah, did this in the beginning and decided to leave it in, can take it out if you're against it, but IMO it's cleaner use of the APIs.

Sources/TSCBasic/Process.swift
sergiocampama: yeah, I think WritableByteStream might be better... we could do that in a later PR, and have OutputByteStream as a type alias to Writable. I prefer that approach to renaming the type as I'm doing here actually, so I'll revert to OutputByteStream

Sources/TSCBasic/Process.swift
sergiocampama: right, which is why I left the `closeOnDeinit` parameter on the `LocalFileOutputByteStream` initializer, to provide a visual cue that that's the expectation.

This should provide API compatibility with current uses of Process, since it just closes the parent side of the pipe. 

Not sure what you mean by the "risk of returning a value that has to be released" since that seems to be a more cumbersome API with more pitfalls than the current proposal.

Sources/TSCBasic/Process.swift
sergiocampama: renamed ProcessStdinStream to WritableByteStream for now, with a note to rename OutputByteStream

sergiocampama:
@swift-ci please test

Sources/TSCBasic/Process.swift
abertelrud: Makes sense.  Until we rename `OutputByteStream` to `WritableByteStream` more globally, do you think it's still worth having a custom alias here or could we leave it as `OutputByteStream` in this API (without an alias)?  I think in QuickHelp it will show up as `WritableByteStream` with at first glance still looks like something special.  So I guess I'm suggesting in the long run renaming `OutputByteStream` to `WritableByteStream` everywhere, but for now keeping it as `OutputByteStream` without an alias.

Sources/TSCBasic/Process.swift
abertelrud: No, I was just curious if I was missing some semantic change.

Sources/TSCBasic/Process.swift
abertelrud: I should have been clearer.  I was trying to point out what I saw as a potential problem, which is that we're now introducing a return value that might hold the file descriptor open longer than expected.  Is the returned stream guaranteed to be deinited immediately for any existing clients that ignore the return value?

Sources/TSCBasic/Process.swift
sergiocampama: sg, using OutputByteStream

Sources/TSCBasic/Process.swift
sergiocampama: not sure about guaranteed, but since there are no references to the `LocalFileOutputByteStream` returned, it does get deinited the moment `launch()` returns, which in turn closes the file, so it's never open longer than required... do you think there are certain scenarios where that won't happen? it seems to me that the guarantee is given by Swift semantics (given that we're not holding a reference to the returned byte stream in the process object itself).

sergiocampama:
@swift-ci please test

Sources/TSCBasic/Process.swift
abertelrud: That makes sense.  It's probably an unfounded concern, from years of having been burned by Objective-C autorelease pools, which made tying any kind of resource deallocation to object lifetime be very perilous.

So I guess it's not a concern for existing clients, which don't capture the return value, and the only possible thing to watch out for would be a client that captures the return value and then waits for the process to finish, which would be a deadlock if the process waits for the last input from stdin.

For example, something like this:
```
let process = Process(arguments: "shasum")
let writer = process.launch()
writer.write(someBytes)
process.waitForExit()
```
Unless we know for sure that `writer` is deinited before the `waitForExit()`, this would hang.

abertelrud: One alternative that I used in a similar implementation (unfortunately not in open source, so I can't point to it) was to have the Process initializer take a block that would be called in the background to write data to the stdin of the process.  The block took a OutputStream as a parameter, and made sure it was closed after the data had been written.  I can see how that can be more constraining for some long-lived processes, though.

MaxDesiatov:
@swift-ci please test

Sources/TSCBasic/Process.swift
sergiocampama: yeah, that's a good point, it sucks not having an API to close the `OutputByteStream` (other than `let stdinStream: OutputByteStream? = process.launch()` and then `stdinStream = nil`)... 

regarding that example on waiting, it's sort of similar to using a process that expects stdin from a shell script... the only difference is that in the script it's signaled with an `EOF`, and in Swift it would be signalled by nil-ling the stdinStream pointer... unless we added API support for EOF/closing to `OutputByteStream`, not sure what your thoughts are on that

abertelrud: Whether or not it's ugly to have an explicit `close()` action in an API like this is a matter of some debate.  My position is that any object that controls an actual resource (other than memory) that has some kind of external semantics, such as a stream, should have some explicit way for a client to say that it's done with that resource.  This is because of the unavoidable side effects of not being able to control the lifetime of the resource.  In other words, it's great that we (for the most part) don't have to think about object lifetimes in Swift, but things like closing a file, stopping listening on a connection, finalizing a database, etc have externally visible semantics and therefore should have explicit control (side note, though: if the user forgets to do it, then deinit should also release the resource automatically).

Interestingly, the file-oriented streams in TSC do have API to close them, we just don't expose it in the generic `OutputStream` that they specialize.  I think we should expose it, to give clients a way to signal intent, even if it's a no-op for some stream implementations (and it's not even always a no-op — for example, even a memory stream could use that as an optimization, by trimming any extraneously allocated memory pages, etc).  Also, having an explicit close action is a way to guarantee that no more information will be written by some other part of the code who might be holding a reference to the stream — that can be important when writing out a file format for example.

So long story short, yes, I do think that we should expose `close` on `OutputByteStream` in general, because in practice it does end up being important, though I know that some people feel it's extraneous.  My own observation is that the more one has been burned by issues around not having explicit control over the end of the stream, the more one tends to want to surface it in the API even though it's slightly unfortunate.  In the end, it ends up being needed.

sergiocampama: yeah, I agree with you. that said, do you want to add the API to OutputByteStream first and then merge this? or is this PR unblocked from that perspective? I'll have a proposal for the OutputByteStream `close` API in a subsequent PR (and another one to rename OutputByteStream to WritableByteStream)

abertelrud: If you need these changes, I wouldn't necessarily block them on reworking the stream.  Maybe you could change the return type to `FileOutputByteStream`, since that's what it always is anyway, which would give callers the `close` function they could use?

abertelrud: It would be leaking the implementation a little bit, but having control over the file descriptor is probably worth it.

sergiocampama: going to do the rename of the stream protocol in https://github.com/apple/swift-tools-support-core/pull/125, and then adding the `close()` API. Seems a cleaner approach to not expose the underlying type, especially since for Windows we're not using `FileOutputByteStream`.

sergiocampama: https://github.com/apple/swift-tools-support-core/pull/125 for the rename PR

abertelrud: Nice, that does seem cleaner.

abertelrud (approved):

sergiocampama:
@swift-ci please test

Sources/TSCBasic/Process.swift
sergiocampama: Updated the PR based on this plan, which was executed on #125  and #127

sergiocampama:
@swift-ci please test

sergiocampama:
@swift-ci please test

sergiocampama:
@abertelrud any other notes before merging this? just making sure I didn't miss anything

PR #123, opened 2020-09-11
[WIP] Add support for Windows and MSVC

stevapple:
- Add `.windows` to `Platform`
- Add `.msvc` to `Triple.ABI`

Discussion & Original PR: apple/swift-package-manager#2920

### DIFF
--- a/Sources/TSCUtility/Platform.swift
+++ b/Sources/TSCUtility/Platform.swift
@@ -26,16 +26,16 @@ public enum Platform: Equatable {
 
     /// Lazily checked current platform.
     public static var currentPlatform = Platform._findCurrentPlatform(localFileSystem)
-    /// Attempt to match `uname` with recognized platforms.
-    /// Additionally, try `ver` to recognize Windows.
+
     public static func _findCurrentPlatform(_ fs: FileSystem) -> Platform? {
-        guard let uname = try? Process.checkNonZeroExit(args: "uname").spm_chomp().lowercased() else {
-            if let ver = try? Process.checkNonZeroExit(args: "ver"),
-                ver.hasPrefix("Microsoft Windows") {
-                return .windows
-            }
-            return nil
-        }
+        /// Recognize Darwin and Windows at compile time.
+      #if os(Darwin)
+        return .darwin
+      #elseif os(Windows)
+        return .windows
+      #else
+        /// Attempt to match `uname` with other recognized platforms.
+        guard let uname = try? Process.checkNonZeroExit(args: "uname").spm_chomp().lowercased() else { return nil }
         switch uname {
         case "darwin":
             return .darwin
@@ -44,6 +44,7 @@ public enum Platform: Equatable {
         default:
             return nil
         }
+      #endif
     }
 
     public static func _findCurrentPlatformLinux(_ fs: FileSystem) -> Platform? {
--- a/Sources/TSCUtility/Platform.swift
+++ b/Sources/TSCUtility/Platform.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+ Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
@@ -16,6 +16,7 @@ public enum Platform: Equatable {
     case android
     case darwin
     case linux(LinuxFlavor)
+    case windows
 
     /// Recognized flavors of linux.
     public enum LinuxFlavor: Equatable {
@@ -26,8 +27,15 @@ public enum Platform: Equatable {
     /// Lazily checked current platform.
     public static var currentPlatform = Platform._findCurrentPlatform(localFileSystem)
     /// Attempt to match `uname` with recognized platforms.
+    /// Additionally, try `ver` to recognize Windows.
     public static func _findCurrentPlatform(_ fs: FileSystem) -> Platform? {
-        guard let uname = try? Process.checkNonZeroExit(args: "uname").spm_chomp().lowercased() else { return nil }
+        guard let uname = try? Process.checkNonZeroExit(args: "uname").spm_chomp().lowercased() else {
+            if let ver = try? Process.checkNonZeroExit(args: "ver"),
+                ver.hasPrefix("Microsoft Windows") {
+                return .windows
+            }
+            return nil
+        }
         switch uname {
         case "darwin":
             return .darwin
--- a/Sources/TSCUtility/Platform.swift
+++ b/Sources/TSCUtility/Platform.swift
@@ -28,11 +28,11 @@ public enum Platform: Equatable {
     public static var currentPlatform = Platform._findCurrentPlatform(localFileSystem)
 
     public static func _findCurrentPlatform(_ fs: FileSystem) -> Platform? {
-        /// Recognize Darwin and Windows at compile time.
-      #if os(Darwin)
-        return .darwin
-      #elseif os(Windows)
+        /// Recognize Windows and Android at compile time.
+      #if os(Windows)
         return .windows
+      #elseif os(Android)
+        return .android
       #else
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci please test

Sources/TSCUtility/Triple.swift
```diff
@@ -102,6 +103,9 @@ public struct Triple: Encodable, Equatable {
         if string.hasPrefix(ABI.android.rawValue) {
             return ABI.android
         }
+        if string == ABI.msvc.rawValue {
```
compnerd, line 106: This should be `string.hasPrefix(ABI.msvc.rawValue)`.  This is a valid triple: `x86_64-unknown-windows-msvc19.26.28806-macho`

Sources/TSCUtility/Platform.swift
```diff
@@ -26,16 +26,16 @@ public enum Platform: Equatable {
...
     public static func _findCurrentPlatform(_ fs: FileSystem) -> Platform? {
-        guard let uname = try? Process.checkNonZeroExit(args: "uname").spm_chomp().lowercased() else {
-            if let ver = try? Process.checkNonZeroExit(args: "ver"),
-                ver.hasPrefix("Microsoft Windows") {
-                return .windows
-            }
-            return nil
-        }
+        /// Recognize Darwin and Windows at compile time.
+      #if os(Darwin)
```
compnerd, line 32: I don't think that is is a valid check - this should be `os(iOS) || os(macOS) || os(tvOS) || os(watchOS)`

compnerd (changes requested):
Conceptually this is fine, what is the visible impact of this change?  (That is, why did this not surface in use of the library?)

Sources/TSCUtility/Platform.swift
stevapple: What about `#if canImport(Darwin)`?

stevapple:
> That is, why did this not surface in use of the library?

I'm going to further its (adding `.windows`) use. Since TSC is now shared by various components, I think it will be useful in some platform-specific implementations.

The `.msvc` change has already reflected in `staticLibraryExtension`, and its important for distinguishing native Windows ABI from those POSIX ports, like what we do to distinguish Android ABI from ordinary Linux.

Sources/TSCUtility/Platform.swift
abertelrud: That might work, though it's an unusual use of `canImport`.  I am not sure where I got the idea that `#if os(Darwin)` was supported (probably from this thread: https://forums.swift.org/t/idea-support-if-os-darwin-as-shorthand-for-os-ios-os-osx-os-watchos-os-tvos) but I was clearly wrong about that.  Thanks for catching that @compnerd.  I'm really not fond of spelling out the Darwin platforms though, so maybe we do need to go back to calling `uname` in this case, unfortunately.

compnerd: @abertelrud you're welcome :)

compnerd:
> > That is, why did this not surface in use of the library?
> 
> I'm going to further its (adding `.windows`) use. Since TSC is now shared by various components, I think it will be useful in some platform-specific implementations.
>

TSC has been in use by multiple components for a while, I'd like to understand what has changed that this suddenly became a problem.  Note that I'm not opposed to the changes, quite the contrary as the existing code is obviously not handling the cases, I'd just like to understand what actually happened so perhaps we can prevent such issues in the future.

> The `.msvc` change has already reflected in `staticLibraryExtension`, and its important for distinguishing native Windows ABI from those POSIX ports, like what we do to distinguish Android ABI from ordinary Linux.

The suffix itself is meaningless, it is purely a convention.  The suffix being `.a` vs `.lib` doesn't distinguish it from being a Windows static library from a Unix one (note that the `.a` suffix is not a POSIX thing, it is a Unix thing - you could say that it is a SuS thing).  In fact, the `.lib` suffix itself can be confusing too - because you don't know whether it is an import library or a static library.  While you are fixing the suffix, please fix the prefix.  I tend to follow the pattern that static libraries have the `lib` prefix and `.lib` suffix but import libraries have only the `.lib` suffix.  This is based on the ucrt/msvcrt model.

An additional question is, do we want to have an additional helper to add the `d` suffix for the debug builds that is the Windows convention?  That is, for a single library you can have the following:

* MD, /MDd:
- liblibrary.lib
- liblibraryd.lib
- library.lib
- libraryd.lib
- library.dll
- libraryd.dll

* /MT, /MTd:
- liblibrary.lib
- liblibraryd.lib
- library.lib
- libraryd.lib
- library.dll
- libraryd.dll

stevapple:
Since the PR is not complete, mark as draft.

I prepare to open other related PRs in packages that rely on TSC and can benefit from this change. This PR is not complete until these related PRs are all done.

stevapple:
> I tend to follow the pattern that static libraries have the lib prefix and .lib suffix but import libraries have only the .lib suffix.

What is the current part of code that generates import libs😂I tried to locate it in `swift-driver` and `swift-package-manager`, but I failed...

abertelrud:
SwiftPM uses llbuild for much of the build machinery, so it's possible that some of the code is there (I don't know how much Windows-specific code it contains, but I've seen some).

compnerd:
The import library is generated via the inferred name by the linker.  It is possible to add an alternate location via `-Xlinker /implies:` which I've found to be extremely important in practice.  Nearly all the swift projects set `CMAKE_OUTPUT_LIBRARY_DIR` and `CMAKE_OUTPUT_RUNTIME_DIR` to collate the libraries and DLLs.  This becomes almost necessary as you cannot run the binaries otherwise as there's no concept if library paths on Windows.

stevapple:
> Nearly all the swift projects set CMAKE_OUTPUT_LIBRARY_DIR and CMAKE_OUTPUT_RUNTIME_DIR to collate the libraries and DLLs.

Is that automatically set by the toolchain or explicitly set by the user? Isn’t it a default behavior to generate the import library right beside the DLL? I need to understand the two questions before I could get into it...

compnerd:
None of that is related to the toolchain, the variables are CMake variables, and under the control of the package and the user.

Sure, the default behavior is to generate it next to it, but the general behaviour is also to put all the binaries next to one another in a single directory, so you are already going to set `CMAKE_OUTPUT_RUNTIME_DIR`.  Might as well as also set `CMAKE_OUTPUT_LIBRARY_DIR`, plus for the case that you have non-runtime libraries (i.e. static libraries) setting the `CMAKE_OUTPUT_ARCHIVE_DIR` and `CMAKE_OUTPUT_LIBRARY_DIR` makes a lot of sense and simplifies a lot of stuff.

Sources/TSCUtility/Platform.swift
```diff
@@ -25,8 +26,15 @@ public enum Platform: Equatable {
 
     /// Lazily checked current platform.
     public static var currentPlatform = Platform._findCurrentPlatform(localFileSystem)
-    /// Attempt to match `uname` with recognized platforms.
+
     public static func _findCurrentPlatform(_ fs: FileSystem) -> Platform? {
+        /// Recognize Windows and Android at compile time.
+      #if os(Windows)
+        return .windows
+      #elseif os(Android)
```
compnerd, line 34: This seems unrelated to the Windows support.  Do you want to split this out?

Sources/TSCUtility/Triple.swift
```diff
@@ -195,7 +199,12 @@ extension Triple {
     
     /// The file extension for static libraries.
     public var staticLibraryExtension: String {
-        return ".a"
+        switch abi {
```
compnerd, line 202: This seems wrong.  It isn't the ABI that dictates the extensions, but rather the platform.  A combination of OS & ABI perhaps, but not just the ABI (e.g., `(.windows, .cygus)` would use `.a`, but `(.windows, .msvc)` would use `.lib`).

compnerd (commented):
I happen to need the windows enumeration constant for the SK-LSP work that I've been doing.

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create pull request with `main` branch.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

PR #131, opened 2020-09-19
TSCBasic: more special case handling for Windows paths

compnerd:
There is no one canonical root on Windows.  Windows contains a forest of
paths, rooted at a drive (`[A-Z]:\\`), a NT Path `\\??\`, or a UNC path,
`\\UNC\` or `\\`.  Avoid the check against a canonical root on Windows.
This is required for `swift package init` to function.


compnerd:
@swift-ci please test

PR #136, opened 2020-10-04
TSCBasic: improve prefix check for Windows

compnerd:
Change the function we use to check for the root path to ensure that we
can properly handle NT paths and UNC paths.


compnerd:
@swift-ci please test

neonichu (approved):

PR #140, opened 2020-10-07
TSCBasic: resolve symbolic links into usable paths

compnerd:
We would provide URI style strings rather than FSR style strings to the
`AbsolutePath` constructor resulting in invalid paths.  This enables
SourceKit-LSP to load settings from the workspace.


compnerd:
@swift-ci please test

abertelrud (approved):
LGTM

PR #141, opened 2020-10-08
build: allow building with Foundation from the toolchain

compnerd:
This is meant to be used to build the Windows toolchain which uses a
staged build where Foundation is not built and used from the build root.


compnerd:
@swift-ci please test

abertelrud (approved):

PR #142, opened 2020-10-09
Revert "build: allow building with Foundation from the toolchain"

brentdax:
Reverts apple/swift-tools-support-core#141, as I suspect it broke TSC tests on Linux.

TSC tests last passed at 11:33 ~~pm~~ am; then we have a couple builds that fail earlier while trying to build the Swift runtime. Starting at 1:55 pm, TSC tests on Linux begin [failing or timing out with many "read error: Bad file descriptor (9)" and "close error: Unknown error -1" messages](https://ci.swift.org/job/swift-PR-Linux-smoke-test/25345/console).

I'm not sure how #141 could have caused these symptoms, but it entered the build during this timeframe and it could have an effect on Linux, and I can't figure out how `Foundation_FOUND` would be set.

### DIFF
--- a/Sources/TSCBasic/CMakeLists.txt
+++ b/Sources/TSCBasic/CMakeLists.txt
@@ -57,10 +57,8 @@ target_compile_options(TSCBasic PUBLIC
 target_link_libraries(TSCBasic PUBLIC
   TSCLibc)
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  if(Foundation_FOUND)
-    target_link_libraries(TSCBasic PUBLIC
-      Foundation)
-  endif()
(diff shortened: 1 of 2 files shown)


brentdax:
@swift-ci test

compnerd:
`Foundation_FOUND` is not expected to be set most of the time.  It's set by `find_package(Foundation CONFIG)`. It has no impact on the package itself.  The library is still linked, just via -lFoundation instead of an absolute path.  It would fail to link if it was the cause.

brentdax:
https://ci.swift.org/job/swift-PR-Linux/24002/console is testing a previously blocked PR with this change to see if it makes a difference.

brentdax:
Probably not it.

PR #144, opened 2020-10-09
TSCBasic: correct `resolveSymlink` on Windows

compnerd:
This would fail to properly resolve the symbolic links on Windows since
the result of the resolution would be a path relative to the location of
the original path.  However, we would instead treat the path as the
absolute path.  Correct this which is required to resolve modulemaps
inside of directories with symbolic links.

This was found by building IndexStoreDB on Windows with SwiftPM.

### DIFF
--- a/Sources/TSCBasic/PathShims.swift
+++ b/Sources/TSCBasic/PathShims.swift
@@ -23,12 +23,12 @@ import Foundation
 /// Returns the "real path" corresponding to `path` by resolving any symbolic links.
 public func resolveSymlinks(_ path: AbsolutePath) -> AbsolutePath {
 #if os(Windows)
-    let resolved: String =
-        (try? FileManager.default.destinationOfSymbolicLink(atPath: path.pathString))
-            ?? path.pathString
+    var resolved: URL = URL(fileURLWithPath: path.pathString)
+    if let destination = try? FileManager.default.destinationOfSymbolicLink(atPath: path.pathString) {
+        resolved = URL(fileURLWithPath: destination, relativeTo: URL(fileURLWithPath: path.pathString))
+    }
 
-    return URL(fileURLWithPath: resolved.standardizingPath)
-              .withUnsafeFileSystemRepresentation {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/TSCBasic/PathShims.swift
```diff
@@ -23,12 +23,12 @@ import Foundation
 /// Returns the "real path" corresponding to `path` by resolving any symbolic links.
 public func resolveSymlinks(_ path: AbsolutePath) -> AbsolutePath {
 #if os(Windows)
-    let resolved: String =
-        (try? FileManager.default.destinationOfSymbolicLink(atPath: path.pathString))
-            ?? path.pathString
+    var resolved: URL = URL(fileURLWithPath: path.pathString)
+    if let destination = try? FileManager.default.destinationOfSymbolicLink(atPath: path.pathString) {
+        resolved = URL(fileURLWithPath: destination, relativeTo: URL(fileURLWithPath: path.pathString))
```
abertelrud, line 28: In theory, could we use `AbsolutePath.init(_ str: String, relativeTo basePath: AbsolutePath)` here, which takes a string such as what `destinationOfSymbolicLink` returns (either relative or absolute path) and returns the resulting absolute path?

abertelrud (approved):
This looks looks good to me, though I think it could be done in a somewhat cleaner way without URL (unless I'm missing something).  But that could be a follow-on PR.

abertelrud:
Also, it would be great to add more testing around this (possibly in separate PR), although I realize that such tests aren't yet running in CI on Windows (I think...?).

Sources/TSCBasic/PathShims.swift
compnerd: We could, but we want the standardized URL ... so we would need to construct the URL first anyway.

compnerd:
Yeah, the tests don't work because the tests explicit create file systems in memory using `/` rather than the `\\` separator which is required on Windows.  Happy to do follow up cleanups to this path though.

PR #145, opened 2020-10-10
Introduce a new visitor-based API for reading LLVM bitstreams

owenv:
This gives clients more manual control over the decoding process. It will allow reading .dia and .swiftdeps files in a single pass, and extracting cross-module dependency info from .swiftmodule files while skipping irrelevant blocks and records. The corresponding swift-driver PR (https://github.com/apple/swift-driver/pull/299) gives a better idea of the applications I have in mind.

The core of this change is the new `BitstreamVisitor` protocol, which has four requirements:
`func validate(signature: Bitcode.Signature) throws`: used to check the signature of a bitstream
`mutating func shouldEnterBlock(id: UInt64) throws -> Bool`: determines if the contents of a block should be read or skipped over
`mutating func didExitBlock() throws`: callback when finished reading a block
`mutating func visit(record: BitcodeElement.Record) throws`: callback whenever a record is read

### DIFF
--- a/Sources/TSCUtility/Bits.swift
+++ b/Sources/TSCUtility/Bits.swift
@@ -84,6 +84,15 @@ struct Bits: RandomAccessCollection {
       return buffer.buffer.dropFirst(offset >> 3).prefix((newOffset - offset) >> 3)
     }
 
+    mutating func skip(bytes count: Int) throws {
+      precondition(count >= 0)
+      precondition(offset & 0b111 == 0)
+      let newOffset = offset &+ (count << 3)
+      precondition(newOffset >= offset)
+      if newOffset > buffer.count { throw Error.bufferOverflow }
+      offset = newOffset
+    }
+
     mutating func advance(toBitAlignment align: Int) throws {
       precondition(align > 0)
       precondition(offset &+ (align&-1) >= offset)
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

abertelrud:
Thanks for adding this; this is a great generalization.  One use of this that I am very much looking forward to is to extend the Diagnostics subsystem in SwiftPM so we can have proper diagnostics from package manifests, rather than have to pass the .dia files up to the clients and rely on them to pass it to libclang.  Thanks for doing the work to parse the bitcode in Swift!

Since the bitstreams are now public types, as of this PR, I wonder if it would be worth also including a doc comment that just has the URL of the description of the bitcode format on llvm.org (or wherever the official version is).  It might help people looking at this for the first time understand the structure.

Thanks!

abertelrud (approved):

DougGregor (approved):
Love it, thanks!

abertelrud:
@owenv Thanks for this; it can be merged any time. Please let me know if you want me to do it.

owenv:
> Since the types related to the bitstream container format are now public types, as of this PR, I wonder if it would be worth also including a doc comment with the URL https://llvm.org/docs/BitCodeFormat.html#bitstream-container-format (or wherever the official version is). It might help people looking at this for the first time understand the structure.

Good suggestion, I linked the reference and added a few more doc comments to some of the public members. I'd appreciate it if you could merge this (assuming tests still pass). Thanks!

owenv:
@swift-ci please test

abertelrud:
Looks great.  Thanks for adding the reference to the documentation!

PR #151, opened 2020-10-27
Fix a unit test (`testBrewPrefix`) that only fails when running all the tests

abertelrud:
This fixes a unit test (`testBrewPrefix`) that only fails when running the whole test suite.  This is because it was dependent on whether package config paths had already been cached.  This adds way to clear the cache.

### DIFF
--- a/Sources/TSCUtility/PkgConfig.swift
+++ b/Sources/TSCUtility/PkgConfig.swift
@@ -33,6 +33,9 @@ public struct PCFileFinder {
     let diagnostics: DiagnosticsEngine
 
     /// Cached results of locations `pkg-config` will search for `.pc` files
+    /// FIXME: This shouldn't use a static variable, since the first lookup
+    /// will cache the result of whatever `brewPrefix` was passed in.  It is
+    /// also not threadsafe.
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please test

mattt (approved):

abertelrud:
@swift-ci please smoke test

abertelrud:
Added some comments to clarify that this isn't so much a fix as a workaround — the problem is that PCFileFinder stores a value in a static variable, so the first lookup "taints" future lookups.  There is a baked-in assumption that the passed-in `brewPrefix` is going to be constant (and indeed that `png-config` won't be installed while the client using PCFileFinder is running).

tomerd (approved):
okay to merge to make sure tests are green, that said we need to track fixing this more holistically as the static caching here is unsafe and incorrect for example if brew changes "under" the process

tomerd:
@swift-ci please test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

PR #154, opened 2020-10-27
Use @_implementationOnly import for TSCclibc

abertelrud:
This change is courtesy of @aciidb0mb3r, who originally opened this as https://github.com/apple/swift-package-manager/pull/2379 against SwiftPM.  Reopening it here as a WIP after discussion with him, because it would be nice to no longer have to copy around header files for clients that use libSwiftPM.dylib and libSwiftPMDataModel.dylib.  TSCclibc is only used by versioning and IndexStore, but the need to have access to the headers extends to all clients.

The IndexStore diffs here get a bit complicated, due to https://bugs.swift.org/browse/SR-11613.  Everything seems to work fine, but we might want to have more discussion about whether there is a better way to get IndexStore to be able to use TSCclibc in an implementation-only manner.

rdar://problem/56219970

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,7 @@
 /*
  This source file is part of the Swift.org open source project
  
- Copyright (c) 2019 Apple Inc. and the Swift project authors
+ Copyright (c) 2019 - 2020 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
  
  See http://swift.org/LICENSE.txt for license information
@@ -45,15 +45,15 @@ let package = Package(
         .target(
             /** Cross-platform access to bare `libc` functionality. */
             name: "TSCLibc",
-            dependencies: ["TSCclibc"]),
+            dependencies: []),
         .target(
             /** TSCBasic support library */
             name: "TSCBasic",
-            dependencies: ["TSCLibc"]),
+            dependencies: ["TSCLibc", "TSCclibc"]),
         .target(
             /** Abstractions for common operations, should migrate to TSCBasic */
             name: "TSCUtility",
-            dependencies: ["TSCBasic"]),
+            dependencies: ["TSCBasic", "TSCclibc"]),
         
         // MARK: Additional Test Dependencies
         
@@ -71,7 +71,7 @@ let package = Package(
         
         .testTarget(
             name: "TSCBasicTests",
-            dependencies: ["TSCTestSupport", "TSCTestSupportExecutable"]),
+            dependencies: ["TSCTestSupport", "TSCTestSupportExecutable", "TSCclibc"]),
         .testTarget(
             name: "TSCBasicPerformanceTests",
             dependencies: ["TSCBasic", "TSCTestSupport"]),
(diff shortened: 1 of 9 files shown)


abertelrud:
@swift-ci please test

abertelrud:
Seems that Process.swift also depends on this on Linux now, so this patch will need to be modified a little.

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

neonichu (approved):

abertelrud:
@compnerd Do the CMake changes here look reasonable to you?  The intent here is to remove the need for having the headers and custom headermap available for all clients of libSwiftPM.

Sources/TSCUtility/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ add_library(TSCUtility
   misc.swift
 )
 target_link_libraries(TSCUtility PUBLIC
+  TSCclibc
```
compnerd, line 47:
```cmake
target_link_libraries(TSCUtiltity PUBLIC
  TSCBasic)
target_link_libraries(TSCUtility PRIVATE
  TSCclibc)
```

The reason to link against `TSCclibc` privately here is that users of the TSCUtility should not need to know about it now that it is `@_implementationOnly`.

compnerd (changes requested):

Sources/TSCUtility/CMakeLists.txt
abertelrud: Thanks, @compnerd.  Pushed changes that implement what I think you're suggesting, which makes total sense, of course.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please test

Package.swift
```diff
@@ -45,15 +45,15 @@ let package = Package(
         .target(
             /** Cross-platform access to bare `libc` functionality. */
             name: "TSCLibc",
-            dependencies: ["TSCclibc"]),
+            dependencies: []),
         .target(
             /** TSCBasic support library */
             name: "TSCBasic",
-            dependencies: ["TSCLibc"]),
+            dependencies: ["TSCLibc", "TSCclibc"]),
```
compnerd, line 52: Hmm, here you have added a dependency on `TSCclibc` for `TSCBasic`, but I don't see that on the CMake side, is that omission intentional?  Is this working around a bug?

abertelrud: Not intentional.  This is because it's needed for some code in Process.swift (which is new since this patch was first written), so it does need to be a dependency.

I was puzzled why I didn't see this in the testing I did using SwiftPM's `bootstrap` script, which builds everything with CMake.  But I just realized that it's using the copy of ToolsSupportCore that's still vendored inside SwiftPM.  Now that I see that I wasn't properly testing these changes in `bootstrap`, I'll make sure it works and update this PR.

Thanks for catching this!

abertelrud: Pushed a fix for this.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please test

abertelrud:
@compnerd Thanks a lot for catching the CMake problems.  Now I have tried building SwiftPM with the changes applied to the vendored TSC copy, I am able to do so successfully.  Do you see any remaining issues with the changes?

compnerd (approved):
LGTM

PR #158, opened 2020-11-03
<rdar://70974921> swift-tsc's AbsolutePath and RelativePath Codable constructors should use validating overloads

jakepetroules:


jakepetroules:
@swift-ci please smoke test

abertelrud (approved):
Thanks @jakepetroules that looks like a great change.

abertelrud:
@swift-ci please test

PR #160, opened 2020-11-05
add closable protocol

tomerd:
motivation: marker protocol for entitites that manage resources and need to be closed to to cleanly free such resources

changes:
* define Closable protocol
* conform WritableByteStrea to Closable
* conform PersistenceCache to Closable

### DIFF
--- a/Sources/TSCUtility/PersistenceCache.swift
+++ b/Sources/TSCUtility/PersistenceCache.swift
@@ -37,7 +37,7 @@ public final class SQLiteBackedPersistentCache: PersistentCacheProtocol, Closabl
     }
 
     deinit {
-        try? db.close()
+        try? self.close()
     }
     
     public func close() throws {
--- a/Sources/TSCUtility/PersistenceCache.swift
+++ b/Sources/TSCUtility/PersistenceCache.swift
@@ -19,7 +19,7 @@ public protocol PersistentCacheProtocol {
 }
 
 /// SQLite backed persistent cache.
-public final class SQLiteBackedPersistentCache: PersistentCacheProtocol {
+public final class SQLiteBackedPersistentCache: PersistentCacheProtocol, Closable {
     let db: SQLite
 
     init(db: SQLite) throws {
@@ -38,6 +38,10 @@ public final class SQLiteBackedPersistentCache: PersistentCacheProtocol {
 
     deinit {
         try? db.close()
+    }
+    
+    public func close() throws {
+        try db.close()
     }
 
     public convenience init(cacheFilePath: AbsolutePath) throws {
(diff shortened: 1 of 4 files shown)


tomerd:
@swift-ci please test

compnerd:
Hmm, what about the case where you have a shared reference that shouldn't be closed?  e.g. you have a non-owning reference to a file descriptor.  Is this supposed to be specific to SQLite?  At least on Windows, close is not universally applicable - some handles are released, some or closed, etc.

Sources/TSCUtility/PersistenceCache.swift
```diff
@@ -39,6 +39,10 @@ public final class SQLiteBackedPersistentCache: PersistentCacheProtocol {
     deinit {
         try? db.close()
```
abertelrud, line 40: I wonder if this should call just `close()` now?  It's a final class, so I guess it doesn't matter, but seems like a more correct pattern in general.

abertelrud: Arguably it should always have called `self.close()` — it's not a consequence of this particular diff, just more cleanup.

abertelrud (approved):
Looks great to me.

Sources/TSCUtility/PersistenceCache.swift
tomerd: good point

tomerd:
> Hmm, what about the case where you have a shared reference that shouldn't be closed? e.g. you have a non-owning reference to a file descriptor. Is this supposed to be specific to SQLite? At least on Windows, close is not universally applicable - some handles are released, some or closed, etc.

Are you referring the name "close" semantically compared to "release"? In the two cases where this applied in this PR the original method was already named `close` so the semantics is preserved.

Is there anything different you want to see here?

tomerd:
@swift-ci please test

compnerd:
Well, I would like to see a bit more documentation on the intent of the protocol and the intended usage.  If the usage is meant to be localised to these two sites, then that name should be fine.  I'm just thinking about what would happen if the protocol ends up being adopted more widely as some platforms don't follow the "everything is a fd" model, and this can raise unnecessary confusion.  The current name is pretty broad and so the documentation would be helpful for a casual reader.

tomerd:
@swift-ci please test

tomerd:
> Well, I would like to see a bit more documentation on the intent of the protocol and the intended usage. If the usage is meant to be localised to these two sites, then that name should be fine. I'm just thinking about what would happen if the protocol ends up being adopted more widely as some platforms don't follow the "everything is a fd" model, and this can raise unnecessary confusion. The current name is pretty broad and so the documentation would be helpful for a casual reader.

got it, thanks for pointing it out. added a comment on intent of API

PR #162, opened 2020-11-06
TSCBasic: repair the Windows build

compnerd:
Due to ambiguity between WinSDK and ucrt's definition of INT_MAX,
explicitly qualify the name.

### DIFF
--- a/Sources/TSCBasic/Lock.swift
+++ b/Sources/TSCBasic/Lock.swift
@@ -96,12 +96,12 @@ public final class FileLock {
         switch type {
         case .exclusive:
             if !LockFileEx(handle, DWORD(LOCKFILE_EXCLUSIVE_LOCK), 0,
-                           DWORD(INT_MAX), DWORD(INT_MAX), &overlapped) {
+                           UInt32.max, UInt32.max, &overlapped) {
                 throw ProcessLockError.unableToAquireLock(errno: Int32(GetLastError()))
             }
         case .shared:
             if !LockFileEx(handle, 0, 0,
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@abertelrud I think that the `TSCclibc` change caused an unintended issue for Windows, this should fix that.

Sources/TSCBasic/Lock.swift
```diff
@@ -96,12 +96,12 @@ public final class FileLock {
         switch type {
         case .exclusive:
             if !LockFileEx(handle, DWORD(LOCKFILE_EXCLUSIVE_LOCK), 0,
-                           DWORD(INT_MAX), DWORD(INT_MAX), &overlapped) {
+                           UInt32.max, UInt32.max, &overlapped) {
```
abertelrud, line 99: Should this be `Int32.max` (since I'm assuming that `INT_MAX` is signed), or does the actual value not matter so much?

abertelrud:
> @abertelrud I think that the TSCclibc change caused an unintended issue for Windows, this should fix that.

Apologies.  I was a bit concerned about that.  How close are we to being able to run `at-swift-ci test` on Windows?

Sources/TSCBasic/Lock.swift
abertelrud: Never mind — I mistakenly thought DWORD was signed.

abertelrud (approved):

PR #152, opened 2020-10-27
Add Symlinks to FileSystem

tgymnich:
I've added symbolic links to `FileSystem` for use in: https://github.com/apple/swift-package-manager/pull/2985

PR #163, opened 2020-11-07
Makes InMemoryFileSystem threadsafe to address flaky unit tests

abertelrud:
Uses a single lock per file system, which results in suboptimal performance but is not a problem in practice:  all of InMemoryFileSystem is unoptimized and is only used in unit tests in practice.

Existing unit tests cover all the functionality, and should make tests such as FileSystemTests.testInMemoryFileSystemFileLock pass reliably now.

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -488,6 +488,9 @@ private class LocalFileSystem: FileSystem {
 //
 /// Concrete FileSystem implementation which simulates an empty disk.
 public class InMemoryFileSystem: FileSystem {
+    
+    /// Private internal representation of a file system node.
+    /// Not threadsafe.
     private class Node {
         /// The actual node data.
         let contents: NodeContents
@@ -501,6 +504,9 @@ public class InMemoryFileSystem: FileSystem {
            return Node(contents.copy())
         }
     }
+
+    /// Private internal representation the contents of a file system node.
+    /// Not threadsafe.
     private enum NodeContents {
         case file(ByteString)
         case directory(DirectoryContents)
@@ -518,6 +524,9 @@ public class InMemoryFileSystem: FileSystem {
             }
         }
     }
+    
+    /// Private internal representation the contents of a directory.
+    /// Not threadsafe.
     private class DirectoryContents {
         var entries: [String: Node]
 
@@ -535,9 +544,16 @@ public class InMemoryFileSystem: FileSystem {
         }
     }
 
-    /// The root filesytem.
+    /// The root node of the filesytem.
     private var root: Node
-    /// Used to access lockFiles in a thread safe manner.
+    
+    /// Protects `root` and everything underneath it.
+    /// FIXME: Using a single lock for this is a performance problem, but in
+    /// reality, the only practical use for InMemoryFileSystem is for unit
+    /// tests.
+    private let lock = Lock()
+    
+    /// Exclusive file system lock vended to clients through `withLock()`.
     private let lockFilesLock = Lock()
 
     public init() {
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please test

abertelrud:
@tgymnich This should fix the flakiness of the test resulting from https://github.com/apple/swift-tools-support-core/pull/152.  It looks as if InMemoryFileSystem was never thread-safe, and so the unit test that uses it was inherently racy.

abertelrud:
Probably best to look at diffs with whitespace disabled — most of this is just a simple implementation wrapping everything with a `withLock`.  In two cases I had to split out the non-locked implementation into separate functions to avoid deadlock where one API function called another API one.  Not going for elegance here, but just to make it robust.

abertelrud:
@tgymnich Not able to add you to list of reviewers for some reason, but would appreciate a look since you've looked at this code more than I have.

abertelrud:
@swift-ci please test

abertelrud:
Doing a bunch of checks after each other since this is a flaky one.  If there's still a problem hopefully one of them will fail.

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

Sources/TSCBasic/FileSystem.swift
```diff
@@ -535,9 +544,16 @@ public class InMemoryFileSystem: FileSystem {
...
-    /// Used to access lockFiles in a thread safe manner.
+    
+    /// Protects `root` and everything underneath it.
+    /// FIXME: Using a single lock for this is a performance problem, but in
+    /// reality, the only practical use for InMemoryFileSystem is for unit
+    /// tests.
+    private let lock = Lock()
+    
+    /// Exclusive file system lock vended to clients through `withLock()`.
     private let lockFilesLock = Lock()
```
tgymnich, line 557: I think we could use replace uses of `lockFilesLock ` with `lock`

abertelrud: Actually, that was the first think I tried — but they really are locking different things.  One is the internal lock for protecting `root`, and the other is the external client lock.  If only one lock is used for this, we have deadlock, since it means that once you've locked the file system, you can't also do any file system operations (`Lock` isn't recursive).

abertelrud: You can see this if you replace use of `lockFilesLock` with `lock` in `withLock()`.  This will cause a couple of the tests to deadlock.

abertelrud: Also, I think with the overall file system lock now in place, I think we can probably restore the ability to do per-file locking as a follow-on PR.

tgymnich (approved):

neonichu (approved):

tomerd (approved):
code looks fine too afaict from a brief glance, adding comment on each method if its expected to be thread safe or not is good idea too. maybe also add a unit test tho that exercises methods in a concurrent fashion, then we can ask @shahmishal to run the tests with TSAN (if that is not already the case) which can help weed out such issues

abertelrud:
Makes sense.  Actually, we have two unit tests that hammer the APIs here concurrently, and those were the ones that were flaky due to race conditions.  This PR should make those unit tests no longer flaky.  I don't think adding another unit tests that tests the concurrency would add anything of value, but having our unit tests run with TSAN enabled is definitely something we should do.  @neonichu is that something that can be done with our own configuration, or would we need Mishal to make changes?

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift please test

PR #164, opened 2020-11-09
Rename "await" to "tsc_await"

DougGregor:
To aid with (further) source compatibility testing, rename `await` to
`tsc_await` to get it out of the way of the use of the `await` keyword
in Swift's Concurrency design. Leave the `await` here as deprecated
entry points so we don't break any code in the process.

Cloned over from https://github.com/apple/swift-package-manager/pull/3036

### DIFF
--- a/Sources/TSCBasic/Await.swift
+++ b/Sources/TSCBasic/Await.swift
@@ -14,11 +14,11 @@
 ///                   should be passed to the async method's completion handler.
 /// - Returns: The value wrapped by the async method's result.
 /// - Throws: The error wrapped by the async method's result
-public func await<T, ErrorType>(_ body: (@escaping (Result<T, ErrorType>) -> Void) -> Void) throws -> T {
-    return try await(body).get()
+public func tsc_await<T, ErrorType>(_ body: (@escaping (Result<T, ErrorType>) -> Void) -> Void) throws -> T {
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

abertelrud (approved):
Do we know who the various clients of TSC are, and what the plan is to update them (besides SwiftPM of course)?  With a lot more activity requiring clients to update (even for good reasons), perhaps it's time to properly version TSC and switch the clients to use the versions.

DougGregor:
> Do we know who the various clients of TSC are, and what the plan is to update them (besides SwiftPM of course)? With a lot more activity requiring clients to update (even for good reasons), perhaps it's time to properly version TSC and switch the clients to use the versions.

That might be reasonable. For this particular case, we can leave the `await` entry point around for clients that aren't going to try to build with concurrency enabled, so we won't break anyone yet... but TSC has enough clients now that proper versioning would likely help us. 

(At the moment, I'm trying to get far enough to build a toolchain for more testing, so TSC + SwiftPM is the target)

abertelrud:
Right, understood, and agreed in this particular case.  There has been at least a couple of cases of revlock between TSC and SwiftPM in the past week, and so I think we should start versioning TSC so that clients can adopt changes at their pace.  I'll start some conversations about that.

PR #166, opened 2020-11-13
Link against missing libraries in TSCUtility CMake

gmittert:
TSUtility uses functions from pthread and dlfcn so it should link
against them to not have undefined symbols.

### DIFF
--- a/Sources/TSCUtility/CMakeLists.txt
+++ b/Sources/TSCUtility/CMakeLists.txt
@@ -46,7 +46,10 @@ add_library(TSCUtility
 target_link_libraries(TSCUtility PUBLIC
   TSCBasic)
 target_link_libraries(TSCUtility PRIVATE
-  TSCclibc)
+  TSCclibc
+  ${CMAKE_DL_LIBS}
+  Threads::Threads)
+
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   target_link_libraries(TSCUtility PRIVATE
     SQLite::SQLite3)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(BUILD_SHARED_LIBS "Build shared libraries by default" YES)
 
 find_package(dispatch QUIET)
 find_package(Foundation QUIET)
+find_package(Threads QUIET)
 find_package(SQLite3 REQUIRED)
 
 add_subdirectory(Sources)


gmittert:
@swift-ci please test platform

Sources/TSCUtility/CMakeLists.txt
```diff
@@ -47,6 +47,13 @@ target_link_libraries(TSCUtility PUBLIC
   TSCBasic)
 target_link_libraries(TSCUtility PRIVATE
   TSCclibc)
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL Darwin)
+target_link_libraries(TSCUtility PRIVATE
+  dl
```
compnerd, line 53: `dl` is not a safe bet - not all platforms require it.  IIRC, bionic doesn't require `libdl` linkage.  `CMAKE_DL_LIBS` is a better option to use here.

`pthread` is also generally not very great.  Using `Threads::Threads` is better via `find_package(Threads)`

compnerd (changes requested):

Sources/TSCUtility/CMakeLists.txt
```diff
@@ -47,6 +47,13 @@ target_link_libraries(TSCUtility PUBLIC
   TSCBasic)
 target_link_libraries(TSCUtility PRIVATE
   TSCclibc)
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL Darwin)
```
abertelrud, line 51: Are the lack of these libraries caused failures on Darwin?  I think the pthread library is brought in through libSystem.

compnerd: No, `pthread` is force linked by the `LC_DYLIB_REEXPORT` of `libSystem_pthread.dylib` from `libSystem`.  This is really a non-Darwin path that it impacts.  I think that something like this is better suited though:

```cmake
find_package(Threads)
target_link_librareis(TSCUtility PRIVATE
  ${CMAKE_DL_LIBS}
  Threads::Threads)
```

Note the fact that this doesn't care about the platform or configuration because that concern is entirely alleviated by offloading that to CMake.

abertelrud: Sounds good.  My question was specifically reacting to the `CMAKE_SYSTEM_NAME STREQUAL Darwin` part of the condition.

gmittert:
Happy to take a platform independent solution. Updated!

gmittert:
@swift-ci please test platform

neonichu (approved):

abertelrud (approved):

gmittert:
I don't have merge permissions for swift-tools-support-core. Can someone merge this for me?

abertelrud:
Looks like merging is blocked because there is still an outstanding request for changes.

abertelrud:
@compnerd Are you fine with the latest edition of the changes?

Sources/TSCUtility/CMakeLists.txt
```diff
@@ -47,6 +47,12 @@ target_link_libraries(TSCUtility PUBLIC
   TSCBasic)
 target_link_libraries(TSCUtility PRIVATE
   TSCclibc)
+
+find_package(Threads)
```
compnerd, line 51: This could possibly be done in the top level CMakeLists.txt

compnerd (approved):
There's some cleanup that could be done, but seems reasonable.  I suppose that it should get tested on windows as well to make sure that Threads::Threads don't need a placeholder as there's no dependent library.

gmittert:
@swift-ci please test platform

gmittert:
@swift-ci please test windows platform

compnerd:
Its local for TSC still :-( (@tomerd might be looking into it); Id like to wait for a snapshot before merging this - I can try it later I suppose as well.

compnerd:
Verified that this seems to be fine on Windows as well - along with the CI passing, I think that this is good to merge.

PR #168, opened 2020-11-16
Fix cachesDirectory using a file URL instead of a path

tgymnich:
I noticed that `LocalFileSystem.cachesDirectory` returns `AbsolutePath("file:///Users/tim/Library/Caches/")`which is not a valid `AbsolutePath`.


abertelrud (approved):
LGTM

MaxDesiatov:
@swift-ci please test

neonichu (approved):

PR #172, opened 2020-12-01
[test] Disable testRerootedFileSystemViewFileLock pending fix

benlangmuir:
This test is racy, so disable until it can be fixed properly.

rdar://71560894


benlangmuir:
@swift-ci please test

abertelrud (approved):

PR #173, opened 2020-12-13
Add some general-purpose path utilities from swift-driver

owenv:
This moves a couple general purpose path utilities from swift-driver to TSC

- array-accepting overloads of `appending(components:)`, to make composition easier
- `basenameWithoutExt` for RelativePath in addition to AbsolutePath


owenv:
@swift-ci please test

neonichu (approved):

abertelrud (approved):

PR #174, opened 2020-12-21
fix error in busy timeout units

tomerd:
motivation: timeout should be in milliseconds but is passed in seconds

cahnges:
* deprecate the SQLite timeout configuration seconds accessor and replace it with a milliseconds one
* add test


tomerd:
@swift-ci please test

neonichu (approved):

PR #175, opened 2020-12-21
Improve finding executable on Windows

stevapple:
There're two main parts of the patch:
- Automatically add `.exe` suffix based on Windows' rule when finding executables;
- Search in Windows-specific paths like `System32` and `Windows`.

Also fixes some problems that prevents tests to compile on Windows, but there's something wrong with my environment so I can only compile them instead of executing them.


Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -87,11 +88,46 @@ class ProcessTests: XCTestCase {
...
+
+            try withCustomEnv(["PATH": tmpdir.pathString]) {
+                XCTAssertNotNil(Process.findExecutable("program.bat"))
+                XCTAssertNil(Process.findExecutable("program.bc"))
+            }
+        }
+      #endif
     }
 
+  #if !os(Windows) // Foundation treats all readable files as executable on Windows.
```
stevapple, line 130: Going to be fixed in apple/swift-corelibs-foundation#2956

Sources/TSCBasic/Process.swift
```diff
@@ -324,9 +325,28 @@ public final class Process: ObjectIdentifierProtocol {
...
+#if os(Windows)
+            var searchPaths = Array<AbsolutePath>()
+            var buffer = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
+
+            // The 32-bit Windows system directory
+            GetSystemDirectoryW(&buffer, .init(MAX_PATH + 1))
+            searchPaths.append(AbsolutePath(String(decodingCString: buffer, as: UTF16.self)))
+
+            // The 16-bit Windows system directory
+            searchPaths.append(AbsolutePath("\(ProcessEnv.vars["systemdrive"] ?? "C:")\\System"))
```
compnerd, line 337: You cannot assume that the system drive is C, you should get this from the system.  If it's unavailable, abort?

Sources/TSCBasic/Process.swift
```diff
@@ -324,9 +325,28 @@ public final class Process: ObjectIdentifierProtocol {
...
+
+            // The Windows directory
+            GetWindowsDirectoryW(&buffer, .init(MAX_PATH + 1))
+            searchPaths.append(AbsolutePath(String(decodingCString: buffer, as: UTF16.self)))
+
+            searchPaths.append(contentsOf: envSearchPaths)
```
compnerd, line 343: Why these specific directories?  Why not the traditional lookup via Path?  What happens when there's a shadowing g variant that is elsewhere in Path and you invoke the wrong executable?  I think that `SearchPathW` would be better.

Sources/TSCBasic/misc.swift
```diff
@@ -87,9 +87,20 @@ public func lookupExecutablePath(
 ) -> AbsolutePath? {
 
     // We should have a value to continue.
+#if os(Windows)
+    guard var value = value, !value.isEmpty else {
+        return nil
+    }
+    let isPath = value.contains("\\")
```
compnerd, line 94: Paths in Win32 may be `/` separated.

Sources/TSCBasic/misc.swift
```diff
@@ -87,9 +87,20 @@ public func lookupExecutablePath(
...
+    if !isPath && !value.contains(".") {
+        value.append(executableFileSuffix)
```
compnerd, line 96: Executable do not need to have the `.exe` suffix - case in point the XCTest bundle.

compnerd (changes requested):

Sources/TSCBasic/misc.swift
stevapple: Executive don’t need to be `.exe`, but they must have a file extension. If there isn’t one, Windows would assume it to be `.exe` (That’s why there’s a`!value.contains(".")` judgement.

Sources/TSCBasic/Process.swift
stevapple: I got the order from https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw . It looks like `%PATH%` has the lowest priority.

Sources/TSCBasic/misc.swift
stevapple: Fixed 👌

Sources/TSCBasic/Process.swift
stevapple: I think we should skip here. The `System` directory in fact is rare for modern Windows machines.

stevapple:
Anyone to push forward this pull request?

abertelrud:
> Anyone to push forward this pull request?

@compnerd Would you be able to take a second look here?  A lot of the swift-tools-support-core contributors don't have much Windows experience.  Thanks!

Sources/TSCBasic/misc.swift
compnerd: I don't think that they need to have an extension.  Use of `CreateProcessW` doesn't care if the image has an extension or not.  I think that the restriction is implemented at the shell level.  Is there documentation that states otherwise?

Tests/TSCBasicTests/ProcessTests.swift
compnerd: This can now be dropped.

Sources/TSCBasic/Process.swift
compnerd: I'm not talking about the system directory, but rather `systemdrive` from the environment.  I'm suggesting something like:

```
guard let systemDrive = ProcessEnv.vars["systemdrive"] else {
  fatalError("required environment variable `systemdrive` is not defined")
}
```

Sources/TSCBasic/Process.swift
compnerd: Correct, the path is added to a list of additional search paths.  That doesn't mean that we shouldn't honor them.

Sources/TSCBasic/misc.swift
stevapple: > If the file name does not contain an extension, .exe is appended. 

From https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

compnerd: > If the file name ends in a period (.) with no extension, or if the file name contains a path, .exe is not appended.

The following sentences are important.  Because we are invoking tools via path, the suffix is not added.

stevapple: Yep, but doesn’t this method means to simulate the path finding process of `CreateProcessW`? That is, when we invoke `lookupExecutablePath("cmd.exe")` we can get `AbsolutePath("C:\\Windows\\System32\\cmd.exe")`, and I think `lookupExecutablePath("cmd")` should have the same output.

This would also help to eliminate platform differences between Windows and POSIX systems.

Tests/TSCBasicTests/ProcessTests.swift
stevapple: Dropped √

Sources/TSCBasic/misc.swift
compnerd: While I agree with the desire to remove the `#if os` checks, I don't agree with the approach.  We already have a way to get the expected extension in a platform agnostic manner, which easily allows us to add the extension without having to conditionalise the code on the OS.  We already know that we expect the files to have the `.exe` suffix.

Alternatively, why not use a number of extensions?  `.cmd` would be a very reasonable wrapper that we might want to use, or perhaps `.ps1`?  But this allows for command injection if there is a site that we missed adding the extension to.

I think I would be against this lookup behaviour change.  Instead, we should audit the sites and use the correct extension.

Sources/TSCBasic/Process.swift
stevapple: Well, I don’t think it’s so necessary...? Swift toolchain can, of course, work without `%systemdrive%`. More importantly, there’re many tests that use blank environment with `PATH` only. Requiring `%systemdrive%` here would break these tests.

Sources/TSCBasic/misc.swift
stevapple: I totally agree that we can use executable file types other than `exe`, but... I don’t think looking for an executable file without extension on Windows is an expected behavior. From the caller’s perspective, one would expect `lookupExecutablePath("clang")` to find `clang.exe` on Windows, instead of finding a `clang` without extension, because both the system API and shell would do so. (And I don’t think we can really invoke Windows scripts from Swift easily, based on the current implementation of `Foundation`?)

Also, `STSC` is a widely-used supporting library, and we need to manually handle the `.exe` suffix platform-specifically nearly everywhere it’s used. That would produce lots of needless codes and may bring more problems. If we adopt the defaulting behavior, hopefully no existing code would break, and anywhere the code breaks there must be a missing suffix in the toolchain. Then we can easily locate and eliminate such problems.

compnerd: I think that if you want to use this to actually find those cases, you should just assert that the path has an extension, which would immediately catch the cases where it does not.  Note that the XCTest bundle does not have an exe suffix, and is currently assumed to be executable.  How are you proposing to handle that?

stevapple: Any executable file type with an extension other than `.exe` will not be affected by this change. That is, you can still use `lookupExecutablePath("TestSomething.xctest")` to find the test bundle, nothing would be different. The only difference is that besides using `lookupExecutablePath("something.exe")`, you can now use `lookupExecutablePath("something")` to retrieve the same result, and that an "executable" without extension would be abandoned.

Sources/TSCBasic/Process.swift
compnerd: IYou should either programatically determine the system drive by querying the registry or require that the environment not be empty.  Any test running with an empty environment is already broken - `INCLUDE` and `LIB` should be set, `AR` should be set as llbuild does not correctly invoke the librarian, and `PATH` is definitely needed to get access to a number of Windows tools (e.g. the resource compiler, the message compiler, etc).

stevapple: I don’t mean to run the test under a blank environment. Instead, some tests needs a custom environment (using `withCustomEnv`) and requiring `%systemdrive%` at runtime will definitely and unnecessarily break those tests. The dedicated test for this function `ProcessTests.testFindExecutable` is among one of them.

stevapple: We shouldn’t require unnecessary environment variables at runtime, which will largely harm compatibility and usability — Even if this library is mainly used in build systems.

tomerd:
@swift-ci please test

Sources/TSCBasic/Process.swift
compnerd: It is not unnecessary environment variables, these are essential environment variables.  `SDKROOT`, `INCLUDE`, `LIB`, `UCRTVersion`, `VCInstallDir`, etc are critical environment variables for the toolchain.  `SystemDrive` IMO falls into the same bucket.  If you want to remove the need for `SystemDrive`, that should be possible, but then it should be queried from the system via the systems API.  We cannot assume `C:` is the system drive.

Sources/TSCBasic/misc.swift
compnerd: You cannot be certain of the use of the API once it is made available.  There is no way to track down each use and ensure that the use satisfies the constraints.  It is better to force the constraints to avoid having issues like that show up later.

Sources/TSCBasic/Process.swift
stevapple: This code clip is outdated. l’ll paste the current implementation here:

```swift
// The 16-bit Windows system directory
if let systemDrive = ProcessEnv.vars["systemdrive"],
   let systemPath = try? AbsolutePath(validating: "\(systemDrive))\\System") {
    searchPaths.append(systemPath)
}
```

The 16-bit SystemDir is actually not available on modern Windows. Keeping this is just to keep align with Windows behavior, and it’s reasonable enough to skip it once  we cannot easily find this directory.

Sources/TSCBasic/misc.swift
stevapple: On the other side, taking this change is enforcing all executables to have a correct extension on Windows. This is exactly the expected behavior. If we find any executable without extension in the toolchain, there should be a bug.

compnerd: No, there shouldn't be any expectation on the file name.  That's the point.

Yes, the toolchain binaries are suffixed with `.exe`.  But this API does not enforce a search within the toolchain.  If you can force TSC to only search within the toolchain directory and nothing else (which means no cmd, no user generated binaries, etc), adding the suffix would be okay.  But even then, an assertion would be better than the forced suffix.

stevapple: According to Microsoft Docs, an executable without extension cannot be executed with its name through `CreateProcess` either. This is nowhere an expected behavior.

If one persists to use such a file name, he should provide a full path here, like what he’s required in `CreateProcess`.

stevapple: As a workaround, we can add (though I think it unnecessary) an optional Boolean parameter here to indicate whether we should automatically add suffixes to it, and defaulting it to `true`.

compnerd: I think that the default should be `false`.  It should be an explicit opt-in.

Sources/TSCBasic/Process.swift
```diff
@@ -324,9 +325,31 @@ public final class Process: ObjectIdentifierProtocol {
                 pathString: ProcessEnv.path,
                 currentWorkingDirectory: localFileSystem.currentWorkingDirectory
             )
+#if os(Windows)
+            var searchPaths = Array<AbsolutePath>()
+            var buffer = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
+
+            // The 32-bit Windows system directory
+            GetSystemDirectoryW(&buffer, .init(MAX_PATH + 1))
```
compnerd, line 333: Please use `buffer.count` here instead of `MAX_PATH + 1`.

Sources/TSCBasic/Process.swift
```diff
@@ -324,9 +325,31 @@ public final class Process: ObjectIdentifierProtocol {
...
+            searchPaths.append(AbsolutePath(String(decodingCString: buffer, as: UTF16.self)))
+
+            // The 16-bit Windows system directory
+            if let systemDrive = ProcessEnv.vars["systemdrive"],
+               let systemPath = try? AbsolutePath(validating: "\(systemDrive))\\System") {
+                searchPaths.append(systemPath)
+            }
+
+            // The Windows directory
+            GetWindowsDirectoryW(&buffer, .init(MAX_PATH + 1))
```
compnerd, line 343: Please use `buffer.count` here instead of `MAX_PATH + 1`.

Sources/TSCBasic/misc.swift
compnerd:
```
.\a.out --stdlib-unittest-in-process --stdlib-unittest-filter "diff param generic => concrete"
```
Note that it uses the traditional Unix spelling (`a.out`) even though the output is not an "a.out" binary.  Expecting `.exe` suffixes for the tools is reasonable, but embedding that into a library I do not think is the correct approach.  Reserving that knowledge to the site of use is better, as there is nothing which mandates the extension.

tomerd:
@compnerd could you review this one please

Sources/TSCBasic/Process.swift
```diff
@@ -350,11 +351,35 @@ public final class Process {
...
+#if os(Windows)
+            var searchPaths = Array<AbsolutePath>()
+            var buffer = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
+
+            // The 32-bit Windows system directory
+            GetSystemDirectoryW(&buffer, .init(MAX_PATH + 1))
+            searchPaths.append(AbsolutePath(String(decodingCString: buffer, as: UTF16.self)))
+
+            // The 16-bit Windows system directory
+            if let systemDrive = ProcessEnv.vars["systemdrive"],
```
compnerd, line 363: Please use `GetSystemDrive` rather than the environment variable.

compnerd (changes requested):
The previously mentioned items are still a concern, I think that this was just rebased.

Sources/TSCBasic/Process.swift
stevapple: I didn’t see such API. `getenv("systemdrive")` seems to be what is commonly used to retrieve the system drive.

stevapple:
I've changed the logic a little bit.

If you search for `clang`, it will now first look for exactly a `clang` in to search paths. If nothing was found, then search for `clang.exe` instead.

compnerd:
> I've changed the logic a little bit.
> 
> If you search for `clang`, it will now first look for exactly a `clang` in to search paths. If nothing was found, then search for `clang.exe` instead.

I've not reviewed the code, but from a design perspective, I think that is something that sounds really good to me.

Sources/TSCBasic/Process.swift
compnerd: Ah, right, sorry about that.  I was thinking of the combination of `GetSystemDirectoryW` and `PathStripToRootW`.

Sources/TSCBasic/misc.swift
```diff
@@ -110,9 +115,14 @@ public func lookupExecutablePath(
     }
 
     // Ensure the value is not a path.
-    if !value.contains("/") {
+    if !isPath {
         // Try to locate in search paths.
         paths.append(contentsOf: searchPaths.map({ $0.appending(component: value) }))
+#if os(Windows)
+        if !value.contains(".") {
```
compnerd, line 122: I think that we should get the last arc and ensure that does not have a `.`.  `\path\name.dir\name` should not be treated as having a `.` in it.

stevapple: This piece of code is wrapped inside `if !isPath`, while `!isPath` indicates that the input `value` is a single component to search for.

Sources/TSCBasic/Process.swift
stevapple: I think I was wrong about the 16-bit system directory. It is placed under Windows directory rather than the root. Now the implementation should not involve any environment varaibles.

stevapple:
`ProcessTests` are still failing either because they’re not tailored for Windows, or because supporting functions (like `withCustomEnv`) are broken. I’d like to address them in another PR.

tomerd:

stevapple:
@tomerd I believe this PR is totally independent of the `System` integration work, and provides the correct behavior on Windows.

It is only blocked by review:(

tomerd:
@compnerd to review

@stevapple aren't there APIs in SwiftSystem that can help reduce boilerplate code here, for example the isPath computation should ideally be something that SystemSystem provides?

stevapple:
> @stevapple aren't there APIs in SwiftSystem that can help reduce boilerplate code here, for example the isPath computation should ideally be something that SystemSystem provides?

If we want to reimplement this check in `FilePath`, then it should look like this:
```swift
if let _ = FilePath.Component(value) {
  // additional logic
}
```

It's somewhat clearer, but shouldn't a blocking issue here. As I've mentioned many times before, this PR is a _bug fix_, while the `FilePath` integration is more of a kind of improvement.  They should have different priorities, and actually have no dependency on each other.

Sources/TSCBasic/Process.swift
```diff
@@ -471,10 +472,26 @@ public final class Process {
...
+            }
+            if GetWindowsDirectoryW(&buffer, .init(buffer.count)) > 0 {
+                let windowsDirectory = String(decodingCString: buffer, as: UTF16.self)
+                // The 16-bit Windows system directory
+                searchPaths.append(AbsolutePath("\(windowsDirectory)\\System"))
+                // The Windows directory
+                searchPaths.append(AbsolutePath(windowsDirectory))
+            }
+#endif
+            searchPaths.append(contentsOf: envSearchPaths)
```
compnerd, line 490: Hmm, I think that this should be inverted no?  We want the environment to take precedent don't we?

Sources/TSCBasic/misc.swift
```diff
@@ -224,6 +224,11 @@ public func lookupExecutablePath(
     guard let value = value, !value.isEmpty else {
         return nil
     }
+#if os(Windows)
+    let isPath = value.contains(":") || value.contains("\\") || value.contains("/")
```
compnerd, line 228: I don't know about the `value.contains(":")`. It should be explicit at index 1 for the drive, or it may be after the file extension as an ADS indicator.

compnerd (commented):
Could you please add more context on the motivation behind this change?

Sources/TSCBasic/misc.swift
stevapple: I will update the naming. This shouldn’t be `isPath`, but `isNotFilename` (will use `isFilename` instead).

Sources/TSCBasic/Process.swift
stevapple: I followed the behavior of `CreateProcess`, where these three directories are always preferred.
But it’s reasonable for TSC to not respect that, though this shouldn’t have any real impact.

compnerd: Hmm, I think that folowing `CreateProcess` is reasonable.

Sources/TSCBasic/misc.swift
```diff
@@ -224,6 +224,11 @@ public func lookupExecutablePath(
     guard let value = value, !value.isEmpty else {
         return nil
     }
+#if os(Windows)
+    let isFileName = !value.contains(":") && !value.contains("\\") && value.contains("/")
```
compnerd, line 228: This seems invalid.  `[/?<>\:*]` are not valid characters for filenames.  `[/\.?*]` at the very least is shared across all the filesystems.  `/` is a path separator, if `value.contains("/")` is true then it is most definitely not a filename.

compnerd (changes requested):

Sources/TSCBasic/misc.swift
stevapple: > `/` is a path separator, if `value.contains("/")` is true then it is most definitely not a filename.

It’s typo here. Thanks for pointing out.

stevapple: > `[/?<>\:*]` are not valid characters for path arcs. `[/\.?*]` at the very least is shared across all the filesystems.

We’re not expecting the others though. If these characters appears, I suppose any resolution could fail.
The main purpose here is to distinguish a file name from any other kinds of _valid_ path. Can you come up with some uncovered cases here?

compnerd: Technically, `\`, `/`, `:`, and `.` can appear in any _valid_ path.  A `.` can appear in a file name.  I think that this is best handled as `PathIsFileSpecW`.

stevapple: AFAIK `PathIsFileSpecW` is equivalent of `{ $0 != nil && !$0.contains(":") && !$0.contains("\\") }`. It didn’t regard `/` as path deliminator.

stevapple: Normalizing the path String can be overhead here, so I would suggest the current implementation.

compnerd: I think that we can live with the overhead for the correctness.  This is meant to be replaced with System in any case.

stevapple: The current implementation is _exactly_ correct and equal. `PathIsFileSpecW` doesn’t change its behavior ever since it was introduced.

Switching to `PathIsFileSpecW` looks a lot uglier and less clear. Also, the API is not actually following what was documented in MSDN, so we needs extra notes on it. That’s unnecessary. Why should we do that??

compnerd: I disagree on that.  I think that it is safer to work with the Win32 APIs rather than trying to re-create it.

stevapple: That's unclear (the API not works as its name suggests);
unsafe (calling C function using an unsafe buffer);
underperforming (constructing buffers and normalizing paths takes a lot more time than iterating the string itself);
unnecessary (Win32 is wrong here, while we can be correct).

stevapple: We cannot rely on Win32 specifically when it comes to paths.

I have investigated on this before I started moving to `System`. Most Win32 APIs work abnormally. They're using very easy implementation that can only fully support FAT paths. We didn't rely on it when we implement `System` either.

stevapple:
I wish we could, but sadly no one has reviewed #219 so far.

It is ready now, I would expect someone to give me some feedback:)

stevapple:
I recently discovered a new bug where Swift Driver and SwiftPM fail to locate themselves. There’re **a bunch of** bugs related to unaligned executable searching logic with Windows system.

I **really** don’t understand why a *bugfix* cannot be merged, and should instead requiring unnecessary refactoring, only meant to replace a small part of its implementation detail? @tomerd @compnerd

PR #178, opened 2021-01-10
make Platform::darwinCacheDirectories thread shafe

tomerd:
motivation: clients like swift-pm use Platform::darwinCacheDirectories in concurrent settings

changes:
* add lock around the internal caching of darwinCacheDirectories
* make several public functions internal since they are only supposed to be exposed for tests
* format code


tomerd:
@swift-ci please test

abertelrud (approved):

PR #179, opened 2021-01-10
add configuration to limit size of sqlite

tomerd:
motivation: there are use cases where users of the SQLite DB may want to limit its size

changes:
* introduce SQLiteCache::Configuraiton allowing to set max database size
* refactor SQLite::checkError to throw more structured errro when database is full, adding infrastructure to other structured errors
* add tests


tomerd:
@swift-ci please test

abertelrud (approved):

PR #182, opened 2021-01-14
Change the current SwiftPM version from 5.3.0 to 5.4.0 (leaving the dev flag set to true)

abertelrud:
ToolsSupportCore used to be part of SwiftPM, and for historical reasons the SwiftPM version is still provided by ToolsSupportCore (this should also be fixed, separately).  In the `main` branch the `dev` flag is still true; this will be set to false in the release branch.

### DIFF
--- a/Sources/TSCUtility/Versioning.swift
+++ b/Sources/TSCUtility/Versioning.swift
@@ -74,7 +74,7 @@ public struct Versioning {
 
     /// The current version of the package manager.
     public static let currentVersion = SwiftVersion(
-        version: (5, 3, 0),
+        version: (5, 4, 0),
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):
we should be moving this to SwiftPM proper

abertelrud:
> we should be moving this to SwiftPM proper

Absolutely.  I filed rdar://73206574 to cover this but that would be more change than we would want at this point in the release.

tomerd:
> Absolutely. I filed rdar://73206574 to cover this but that would be more change than we would want at this point in the release.

agreed

PR #183, opened 2021-01-14
[5.4] Change the current SwiftPM version from 5.3.0 to 5.4.0

abertelrud:
For historical reasons the SwiftPM version number is still contained in ToolsSupportCore (which should be fixed).  This increments it to 5.4.0, but leaves the `dev` flag true for now so that the 999.0 tools version is still accepted until the change has propagated through.  The `dev` flag will then be set to false before 5.4 is released.

This affects what `swift package --version` shows and also allows v5.4 package manifests to be used in SwiftPM.

rdar://73206582

### DIFF
--- a/Sources/TSCUtility/Versioning.swift
+++ b/Sources/TSCUtility/Versioning.swift
@@ -74,7 +74,7 @@ public struct Versioning {
 
     /// The current version of the package manager.
     public static let currentVersion = SwiftVersion(
-        version: (5, 3, 0),
+        version: (5, 4, 0),
         isDevelopment: true,
         buildIdentifier: getBuildIdentifier())
 


abertelrud:
@swift-ci please test

abertelrud:
Scope of Issue:  SwiftPM needs to have its version set to 5.4 so that it can correctly report its version number in `--version` and so that it will accept packages with a tools version of 5.4 and earlier.

Risk:  Low. While this change is in ToolsSupportCore, it does not affect SwiftDriver or SourceKitLSP (there are no mentions of `currentVersion`). There is no logic change, and the usage pattern in SwiftPM is unchanged compared with previous releases.

Testing:  Automated testing through unit tests.  Manual testing in conjunction with SwiftPM.

neonichu (approved):

abertelrud:
cc: @airspeedswift

abertelrud:
Test failure seems to be because https://github.com/apple/swift-tools-support-core/pull/183 hasn't landed.

abertelrud:
I could avoid the failures by setting the `dev` flag back to `true`, but the right thing to do is to get the swiftpm test fixture change in the other PR merged first, since this is a release branch.

abertelrud:
I think to break the cycle here, that's the least bad thing to do.  Leaving at `dev` will still allow `999` to be accepted for now.  We can easily change that in a follow-on diff.

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test macOS

abertelrud:
Merging after approval in rdar://73206582

PR #185, opened 2021-01-21
Mark 'release/5.4' as being a release build

abertelrud:
This makes the `--version` output not add a `-dev` suffix, and also makes the tools version no longer accept `999.0` (`.vNext`) as a valid version (that's the placeholder used during development until a version number is known).

<rdar://73433200>


abertelrud:
@swift-ci please test

tomerd (approved):

PR #187, opened 2021-01-26
Add a BitstreamWriter Abstraction

CodaFi:
Add a counterpart to `BitstreamReader` and shuffle the deck chairs a bit so the different abstractions are well separated and their shared infrastructure is in one place. The first commit largely does the shuffling parts, the latter two commits contain the juicy bitstream writer bits.

rdar://73517080

### DIFF
--- a/Sources/TSCUtility/BitstreamWriter.swift
+++ b/Sources/TSCUtility/BitstreamWriter.swift
@@ -1,0 +1,566 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+public final class BitstreamWriter {
+    public enum BlockInfoCode: UInt8 {
+        case setBID = 1
+        case blockName = 2
+        case setRecordName = 3
+    }
+
+    /// The buffer of data being written to.
+    private(set) public var data: [UInt8]
+
+    /// The current value. Only bits < currentBit are valid.
+    private var currentValue: UInt32 = 0
+
+    /// Always between 0 and 31 inclusive, specifies the next bit to use.
+    private var currentBit: UInt8 = 0
+
+    /// The bit width used for abbreviated codes.
+    private var codeBitWidth: UInt8
+
+    /// The list of defined abbreviations.
+    private var currentAbbreviations = [Bitstream.Abbreviation]()
+
+    /// Represents an in-flight block currently being emitted.
+    struct Block {
+        /// The code width before we started emitting this block.
+        let previousCodeWidth: UInt8
+
+        /// The index into the data buffer where this block's length placeholder
+        /// lives.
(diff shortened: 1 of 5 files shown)


CodaFi:
@owenv As somebody with more knowledge of bitstream than most, I'd like your opinion here as well.

CodaFi:
@swift-ci test

Sources/TSCUtility/BitstreamWriter.swift
```diff
@@ -0,0 +1,665 @@
...
+///     }
+///
+///     // Define the 'version' record and register its name
+///     recordWriter.writeRecord(BitstreamWriter.BlockInfoCode.setRecordName) {
+///         $0.append(DiagnosticRecordID.version)
+///         $0.append("Version")
+///     }
+///
+///     versionAbbrev = recordWriter.defineBlockInfoAbbreviation(.metadata, .init([
+///         .literalCode(RoundTripRecordID.version),
```
CodaFi, line 72: Heh, serves me right for copying and pasting from the test suite.

CodaFi:
I'd be curious if we could provide a less... error-prone API surface than the existing `append`-based one. The C++ uses the `BCRecordLayout` to ensure that e.g. if you mess up the record emission step you get a horrific spew of template errors.

owenv (approved):
Everything here looks good to me!

> I'd be curious if we could provide a less... error-prone API surface than the existing append-based one. The C++ uses the BCRecordLayout type to ensure that e.g. if you mess up the record emission step you get a horrific spew of template errors.

Without variadic generics I think static enforcement of record layouts is pretty tricky. If we want better dynamic enforcement, maybe abbreviations could be types conforming to a protocol like:
```
protocol Abbreviation {
        static var blockID: BlockID
	static var operands: [AbbreviationOperand] { get }
}
```
and then an `AbbreviatedRecordBuffer<AbbreviationType: Abbreviation>` type could have similar `append` methods but with some additional assertions using the extra info. I can't decide if the extra complexity would be worth it though.

CodaFi:
My ideal API would be gobs of magic to bind everything to Codable. But bitstream's abbreviations make that awkward on top of the fact that it's somewhat to know when to use a new block section instead of extending a logical record that way. Another alternative could be a SwiftUI-style DSL to define the schema which would yield templates for the record types that we could check (possibly at runtime). I don't know, it's all kind of messy...

Tests/TSCUtilityTests/BitstreamTests.swift
```diff
@@ -183,4 +183,410 @@ final class BitstreamTests: XCTestCase {
...
+
+            abbreviations.append(recordWriter.defineBlockInfoAbbreviation(.diagnostics, .init([
+                .literalCode(DiagnosticRecordID.fixIt)
+            ] + sourceLocationOperands + sourceLocationOperands + [
+                .fixed(bitWidth: 16), // Text size
+                .blob                // FixIt Text
+            ])))
+        }
+
+        XCTAssertEqual(recordWriter.data, [
```
harlanhaskins, line 518: We could probably pull this test out into its own clang serializer, honestly. Later, not for this PR.

harlanhaskins (approved):
Thank you so much for cleaning up the writer and unifying the abstractions! This looks really awesome.

CodaFi:
@swift-ci test

CodaFi:
⛵

PR #189, opened 2021-02-08
Switch BitstreamReader Entrypoints to Use ByteString

CodaFi:
ByteString is a more appropriate vocabulary type than Data to use for
these entrypoints. TSCBasic traffics in ByteString values, and its
representation as an array means that there are reasonable indexing
invariants we can rely on with respect to subsequences.

This should also enable some clean up in the swift-driver.

### DIFF
--- a/Sources/TSCUtility/Bits.swift
+++ b/Sources/TSCUtility/Bits.swift
@@ -9,15 +9,16 @@
 */
 
 import Foundation
+import TSCBasic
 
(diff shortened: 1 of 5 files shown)


CodaFi:
@swift-ci test

abertelrud (approved):
LGTM.  Only concern would be performance — I don't know how optimized Data is internally compared with ByteString.  But being able to return slices sounds as if it would be a performance win.

PR #190, opened 2021-02-23
Process: only close the read end of the pipe after reading the full content successfully

nkcsgexi:
If we close the read end of the pipe after hitting an error (EIO for example),
the child process may be still running and will print more content into the write end of the
pipe. Such write to a closed-read pipe will cause a SIGPIPE signal and fail.

Potential a fix to rdar://74058113

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -611,6 +611,14 @@ public final class Process: ObjectIdentifierProtocol {
                     break loop
                 }
             case 0:
+                // Close the read end of the output pipe.
+                // We should avoid closing the read end of the pipe in case
+                // -1 because the child process may still have content to be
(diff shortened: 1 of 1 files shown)


nkcsgexi:
@swift-ci please test

tomerd:
seems reasonable to me. @abertelrud @neonichu ?

abertelrud (approved):
Looks like a great find.  This fix looks correct to me.  Thanks!

abertelrud:
But we do end up leaking a file descriptor in this case if we get an error.  Not sure how to remedy that without triggering the `SIGPIPE` though.

neonichu (approved):

nkcsgexi:
Guess leaving a pipe read end unclosed is fine though? The OS will clean it up eventually.

nkcsgexi:
🚢

PR #191, opened 2021-02-23
Improve path normalization and judgment on Windows

stevapple:
- Fixes the problem of `..` being ignored during normalizing;
- Avoids crash when validating an empty path.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -458,8 +458,8 @@ private struct UNIXPath: Path {
             PathCchRemoveFileSpec(data, path.count)
             return String(decodingCString: data, as: UTF16.self)
         }
-        // These two expressions represent for the current directory.
-        if dir == "\\" || dir == "" {
+        // Blank path represents for the current directory.
+        if dir == "" {
             return "."
         }
         return dir
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -620,21 +620,14 @@ private struct UNIXPath: Path {
     }
 
     init(normalizingRelativePath path: String) {
+        let pathSeparator: Character
       #if os(Windows)
-        var result: PWSTR?
-        defer { LocalFree(result) }
-
-        _ = path.replacingOccurrences(of: "/", with: "\\").withCString(encodedAs: UTF16.self) {
-            PathAllocCanonicalize($0, ULONG(PATHCCH_ALLOW_LONG_PATHS.rawValue), &result)
-        }
-
-        var canonicalized: String = String(decodingCString: result!, as: UTF16.self)
-        if canonicalized == "" || canonicalized == "\\" {
-            canonicalized = "."
-        }
-        self.init(string: canonicalized)
+        pathSeparator = "\\"
+        let path = path.replacingOccurrences(of: "/", with: "\\")
       #else
-        precondition(path.first != "/")
+        pathSeparator = "/"
+      #endif
+        precondition(path.first != pathSeparator)
 
         // FIXME: Here we should also keep track of whether anything actually has
         // to be changed in the string, and if not, just return the existing one.
@@ -644,7 +637,7 @@ private struct UNIXPath: Path {
         // the normalized string representation.
         var parts: [String] = []
         var capacity = 0
-        for part in path.split(separator: "/") {
+        for part in path.split(separator: pathSeparator) {
             switch part.count {
             case 0:
                 // Ignore empty path components.
@@ -683,7 +676,7 @@ private struct UNIXPath: Path {
         if let first = iter.next() {
             result.append(contentsOf: first)
             while let next = iter.next() {
-                result.append("/")
+                result.append(pathSeparator)
                 result.append(contentsOf: next)
             }
         }
@@ -694,7 +687,6 @@ private struct UNIXPath: Path {
 
         // If the result is empty, return `.`, otherwise we return it as a string.
         self.init(string: result.isEmpty ? "." : result)
-      #endif
     }
 
     init(validatingAbsolutePath path: String) throws {
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -449,12 +449,15 @@ private struct UNIXPath: Path {
 
     var dirname: String {
 #if os(Windows)
+        guard string != "" else {
+            return "."
+        }
         let fsr: UnsafePointer<Int8> = string.fileSystemRepresentation
         defer { fsr.deallocate() }
 
         let path: String = String(cString: fsr)
         return path.withCString(encodedAs: UTF16.self) {
-            let data = UnsafeMutablePointer(mutating: $0)
+            let data = UnsafeMutaßlePointer(mutating: $0)
             PathCchRemoveFileSpec(data, path.count)
             return String(decodingCString: data, as: UTF16.self)
         }
@@ -685,6 +688,9 @@ private struct UNIXPath: Path {
 
     init(validatingAbsolutePath path: String) throws {
       #if os(Windows)
+        guard path != "" else {
+            throw PathValidationError.invalidAbsolutePath(path)
+        }
         let fsr: UnsafePointer<Int8> = path.fileSystemRepresentation
         defer { fsr.deallocate() }
 
@@ -707,6 +713,9 @@ private struct UNIXPath: Path {
 
     init(validatingRelativePath path: String) throws {
       #if os(Windows)
+        guard path != "" else {
+            self.init(normalizingRelativePath: path)
+        }
         let fsr: UnsafePointer<Int8> = path.fileSystemRepresentation
         defer { fsr.deallocate() }
 
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -12,11 +12,7 @@ import Foundation
 import WinSDK
 #endif
 
-#if os(Windows)
 private typealias PathImpl = UNIXPath
-#else
-private typealias PathImpl = UNIXPath
-#endif
 
 /// Represents an absolute file system path, independently of what (or whether
 /// anything at all) exists at that path in the file system at any given time.
@@ -453,16 +449,12 @@ private struct UNIXPath: Path {
         defer { fsr.deallocate() }
 
         let path: String = String(cString: fsr)
-        let dir: String = path.withCString(encodedAs: UTF16.self) {
+        let result: String = path.withCString(encodedAs: UTF16.self) {
             let data = UnsafeMutablePointer(mutating: $0)
             PathCchRemoveFileSpec(data, path.count)
             return String(decodingCString: data, as: UTF16.self)
         }
-        // Blank path represents for the current directory.
-        if dir == "" {
-            return "."
-        }
-        return dir
+        return result.isEmpty ? "." : result
 #else
         // FIXME: This method seems too complicated; it should be simplified,
         //        if possible, and certainly optimized (using UTF8View).
@@ -621,12 +613,12 @@ private struct UNIXPath: Path {
 
     init(normalizingRelativePath path: String) {
         let pathSeparator: Character
-      #if os(Windows)
+#if os(Windows)
         pathSeparator = "\\"
         let path = path.replacingOccurrences(of: "/", with: "\\")
-      #else
+#else
         pathSeparator = "/"
-      #endif
+#endif
         precondition(path.first != pathSeparator)
 
         // FIXME: Here we should also keep track of whether anything actually has
@@ -690,7 +682,7 @@ private struct UNIXPath: Path {
     }
 
     init(validatingAbsolutePath path: String) throws {
-      #if os(Windows)
+#if os(Windows)
         guard path != "" else {
             throw PathValidationError.invalidAbsolutePath(path)
         }
@@ -702,7 +694,7 @@ private struct UNIXPath: Path {
             throw PathValidationError.invalidAbsolutePath(path)
         }
         self.init(normalizingAbsolutePath: path)
-      #else
+#else
         switch path.first {
         case "/":
             self.init(normalizingAbsolutePath: path)
@@ -711,11 +703,11 @@ private struct UNIXPath: Path {
         default:
             throw PathValidationError.invalidAbsolutePath(path)
         }
-      #endif
+#endif
     }
 
     init(validatingRelativePath path: String) throws {
-      #if os(Windows)
+#if os(Windows)
         guard path != "" else {
             self.init(normalizingRelativePath: path)
             return
@@ -728,14 +720,14 @@ private struct UNIXPath: Path {
             throw PathValidationError.invalidRelativePath(path)
         }
         self.init(normalizingRelativePath: path)
-      #else
+#else
         switch path.first {
         case "/", "~":
             throw PathValidationError.invalidRelativePath(path)
         default:
             self.init(normalizingRelativePath: path)
         }
-      #endif
+#endif
     }
 
     func suffix(withDot: Bool) -> String? {
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -431,7 +431,11 @@ extension Path {
 private struct UNIXPath: Path {
     let string: String
 
+#if os(Windows)
+    static let root = UNIXPath(string: "\\")
+#else
     static let root = UNIXPath(string: "/")
+#endif
 
     static func isValidComponent(_ name: String) -> Bool {
 #if os(Windows)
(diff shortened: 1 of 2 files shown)


tomerd:
@swift-ci test

tomerd:
@compnerd

Sources/TSCBasic/Path.swift
```diff
@@ -544,11 +549,13 @@ private struct UNIXPath: Path {
...
+        var result: PWSTR?
+        defer { LocalFree(result) }
+
         _ = path.withCString(encodedAs: UTF16.self) {
-            PathCanonicalizeW(&buffer, $0)
+            PathAllocCanonicalize($0, ULONG(PATHCCH_ALLOW_LONG_PATHS.rawValue), &result)
         }
-        self.init(string: String(decodingCString: buffer, as: UTF16.self))
+        self.init(string: String(decodingCString: result!, as: UTF16.self))
       #else
```
compnerd, line 559: If you want to change this, I think it is better to change this to `PathCchCanonicalize` rather than using `PathAllocCanonicalize`.  I assume that the reason for this change is the long path support?

compnerd (changes requested):

Sources/TSCBasic/Path.swift
```diff
@@ -707,6 +724,10 @@ private struct UNIXPath: Path {
 
     init(validatingRelativePath path: String) throws {
       #if os(Windows)
+        guard path != "" else {
+            self.init(normalizingRelativePath: path)
+            return
+        }
```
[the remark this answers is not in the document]
stevapple: "" is always a valid relative path. I don’t think there’s any changes. Adding the judgement here simply because retrieving `fileSystemRepresentation` of a blank string will halt the whole program.

Sources/TSCBasic/Path.swift
stevapple: Both seems okay to me, so I picked `PathAllocCanonicalize` for long path support. Deprecating `PathCanonicalizeW` for memory safety.

Sources/TSCBasic/Path.swift
```diff
@@ -453,11 +453,16 @@ private struct UNIXPath: Path {
...
 
         let path: String = String(cString: fsr)
-        return path.withCString(encodedAs: UTF16.self) {
+        let dir: String = path.withCString(encodedAs: UTF16.self) {
             let data = UnsafeMutablePointer(mutating: $0)
             PathCchRemoveFileSpec(data, path.count)
             return String(decodingCString: data, as: UTF16.self)
         }
+        // These two expressions represent for the current directory.
+        if dir == "\\" || dir == "" {
```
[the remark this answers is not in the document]
stevapple: Just pass "main.swift" and we’ll get "" :(
I double checked and find "\\" not the case here, fixed

Sources/TSCBasic/Path.swift
```diff
@@ -614,11 +621,18 @@ private struct UNIXPath: Path {
 
     init(normalizingRelativePath path: String) {
       #if os(Windows)
-        var buffer: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
+        var result: PWSTR?
+        defer { LocalFree(result) }
+
         _ = path.replacingOccurrences(of: "/", with: "\\").withCString(encodedAs: UTF16.self) {
-            PathCanonicalizeW(&buffer, $0)
+            PathAllocCanonicalize($0, ULONG(PATHCCH_ALLOW_LONG_PATHS.rawValue), &result)
```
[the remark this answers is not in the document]
stevapple: CC @compnerd

Sources/TSCBasic/Path.swift
compnerd: PathCch APIs are recommended.  I don't see what memory safety issues it avoids.  Yes, I get that it is a mutating API, but it is actually safer - it doesn't do additional allocations.

Sources/TSCBasic/Path.swift
```diff
@@ -613,14 +612,14 @@ private struct UNIXPath: Path {
...
-      #else
-        precondition(path.first != "/")
+        let pathSeparator: Character
+#if os(Windows)
+        pathSeparator = "\\"
+        let path = path.replacingOccurrences(of: "/", with: "\\")
+#else
+        pathSeparator = "/"
+#endif
+        precondition(path.first != pathSeparator)
```
compnerd, line 622: This precondition doesn't hold.  `\path` is valid as a relative path.  It is relative to the current drive.

Sources/TSCBasic/Path.swift
```diff
@@ -680,11 +679,13 @@ private struct UNIXPath: Path {
 
         // If the result is empty, return `.`, otherwise we return it as a string.
         self.init(string: result.isEmpty ? "." : result)
```
compnerd, line 682: Removing the use of `PathCanonicalizeW` is fine, but this no longer canonicalizes all paths.  Please handle UNC paths and drive roots.

Sources/TSCBasic/Path.swift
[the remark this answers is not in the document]
compnerd: Ah, I see.  This deserves a comment I think.

compnerd (changes requested):

Sources/TSCBasic/Path.swift
stevapple: Although it is literally "relative", Windows Shell (and Swift which depends on it) regards `\path` `\?\path` `\\share\path` as absolute paths (according to the response of `PathIsRelativeW`). Personally I think these paths should be treated the same with a regular absolute path. For example, `\path\..\..` indicates `\` like `C:\path\..\..` indicates `C:\`, but `path\..\..` indicates `..`.

Sources/TSCBasic/Path.swift
```diff
@@ -544,11 +546,13 @@ private struct UNIXPath: Path {
 
     init(normalizingAbsolutePath path: String) {
       #if os(Windows)
-        var buffer: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
-        _ = path.withCString(encodedAs: UTF16.self) {
-            PathCanonicalizeW(&buffer, $0)
+        var result: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
+        defer { LocalFree(result) }
```
compnerd, line 550: The `LocalFree` here is incorrect.  The allocation will be released by the `deinit` of `Array` which will be correct.

Sources/TSCBasic/Path.swift
```diff
@@ -941,3 +955,11 @@ private func mayNeedNormalization(absolute string: String) -> Bool {
     }
     return false
 }
+
+#if os(Windows)
+fileprivate extension String {
+    func prenormalized() -> String {
```
compnerd, line 961: Would you mind renaming this to `standardizingSeparator` please?

Sources/TSCBasic/Path.swift
compnerd: UNC paths are always absolute.  It is just a drive relative path which is not absolute.  That distinction is important as it impacts the lexical traversal of the path string.

compnerd (changes requested):

Sources/TSCBasic/Path.swift
stevapple: `PathIsRelativeW` regards `\path` as an absolute path. Also, if we treat it as relative, behavior regarding `..` will be unaligned.

stevapple: Ping @compnerd

Sources/TSCBasic/Path.swift
```diff
@@ -544,11 +547,12 @@ private struct UNIXPath: Path {
...
     init(normalizingAbsolutePath path: String) {
       #if os(Windows)
-        var buffer: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
-        _ = path.withCString(encodedAs: UTF16.self) {
-            PathCanonicalizeW(&buffer, $0)
+        var result: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
+
+        _ = path.standardizingPathSeparator().withCString(encodedAs: UTF16.self) {
+            PathCchCanonicalize($0, result.length, $0)
         }
```
compnerd, line 554: I'm confused.  How does this work?  `result` is unused other than to compute its length?  This will just mutate the string in place, which would result in the result always being empty?

Sources/TSCBasic/Path.swift
compnerd: I still think that the differences are material and need to be addressed.  Paths in Windows do not behave like Unix, and so you will have differences.  e.g.

path1 = "C:\Users\compnerd"
path2 = "D:"
path1 + path2 is not `C:\Users\compnerd\D:" but rather "C:\Windows\System32\WindowsMetadata\"

because the current directory on D: happened to be "..\..\Windows\System32\WindowsMetadata"

compnerd (changes requested):

Sources/TSCBasic/Path.swift
stevapple: Thanks for pointing out the bug! Fixed.

Sources/TSCBasic/Path.swift
stevapple: I’ll look into more critical cases here. 

However, if I regard `\path` as a relative path, it will change the behavior of a bunch of APIs (while this PR focuses on bug fixes), so that change definitely needs to be addressed in a separate PR.

stevapple: IMO I believe this PR is fully capable of fixing faulty behavior without changing the current judgement of relative paths and absolute paths. To fully support Windows path system, we’d better work on another one which may contain a complete refactoring.

What this PR mainly addressed and fixed is the problem of `..` being ignored, and of crashing on an empty path, which are critical bugs for tools like SwiftPM.

stevapple:
If possible, I’d like to refactor it using `swift-system` after FilePath API is accepted. I don’t see the need to handle those tricky cases repeatedly in separate packages.

tomerd:
> If possible, I’d like to completely refactor Path using swift-system after FilePath API is accepted. I don’t see the need to handle those tricky cases repeatedly in separate packages.

@stevapple we'd love to see PRs to that end, its definitely the right direction

stevapple:
CC @compnerd 

Regardless of any future improvements, the bug fixed in this PR is critical for correcting SwiftPM behavior on Swift 5.4 and later. I’d like to see it merged into `release/5.4` before the official release.

stevapple:
@tomerd Could you merge this or touch @compnerd ? He seems out of track of this PR.
But it’s critical for SwiftPM on Windows in Swift 5.4. We’re running out of time.

compnerd:
I think that this change is way too far for the 5.4 release.

I think that it is important to address the issues with the model though.

stevapple:
I’d expect `System` to bring the fully-functional path parsing.

Shipping a SwiftPM which cannot resolve `.package(path: "../AnotherPackage")` with 5.4 is definitely not what we’d like to see.

stevapple:
@ddunbar May you take a look at this PR? :) I believe that's a critical one for the 5.4 release.

Also @tomerd here because this mainly solves a bug in SwiftPM.

tomerd:
> I’d expect System to bring the fully-functional path parsing.

I propose a phased approach:

First, please work with @compnerd to get this PR to a mergeable state. 

Later, we should start landing PRs that refactor TSC to use SwiftSystem.

> Shipping a SwiftPM which cannot resolve .package(path: "../AnotherPackage") with 5.4 is definitely not what we’d like to see.

its too late for 5.4.0, but once we merge this PR into main, we should open a PR against the 5.4 branch so that **if/when** we roll a 5.4.1 dot release we can include it

stevapple:
> First, please work with @compnerd to get this PR to a mergeable state.

I’m just missing the point to evolve this PR. I believe in its current state it’s making things better (but not perfect yet) without introducing new problems.

@compnerd Can you make a detailed review on the codes?

stevapple:
Any future directions? @compnerd

compnerd:
I don't think that all the previous comments on the change have been addressed yet.

compnerd:
I'm not sure that this is the right way to handle that.  It just behaves differently after this change.  That doesn't make it more robust.  I agree with you that this is in need of a rewrite - and really, this seems like the ideal case and time to migrate the path handling to Swift System.  The 5.5 release branch is already in slushy freeze, so we have plenty of time for the next release to get the implementation migrated and corrected.

stevapple:
> It just behaves differently after this change.

I would like to know which cases this PR breaks...  I think we'd better get a workaround for 5.4, and then get to rewrite it with Swift System for the 5.5 release.

compnerd:
> I would like to know which cases this PR breaks... I think we'd better get a workaround for 5.4, and then get to rewrite it with Swift System for the 5.5 release.

It already is too late for all of those.  Such a large change for 5.4 is entirely out of the question - it is already finalized.  I think that such a change for 5.5 is unreasonable.  It is too far in the cycle to make that type of change.  This would be best slated for the next release for which we have plenty of time to do by means of replacing the path handling with Swift System.

stevapple:
> It already is too late for all of those. Such a large change for 5.4 is entirely out of the question - it is already finalized.

Any critical bug fix should be allowed. I don’t see this PR introducing new behavior AFAK (as for the result) and the bugs of old implementation are leading to unexpected crashes and extremely strange behavior.

stevapple:
```swift
import WinSDK

let path = CommandLine.argc > 1 ? CommandLine.arguments[1] : "..\\swift-tools-support-core"
print("user input: " + path)
var buffer: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
_ = path.withCString(encodedAs: UTF16.self) { PathCanonicalizeW(&buffer, $0) }
print("normalized (current): " + String(decodingCString: buffer, as: UTF16.self))

var parts: [String] = []
var capacity = 0
for part in path.split(separator: "\\") {
    switch part.count {
    case 0: continue
    case 1 where part.first == ".": continue
    case 2 where part.first == "." && part.last == ".":
        guard let prev = parts.last else { fallthrough }
        if !(prev.count == 2 && prev.first == "." && prev.last == ".") {
            parts.removeLast()
            capacity -= prev.count
            continue
        }
        fallthrough
    default:
        parts.append(String(part))
        capacity += part.count
    }
}
capacity += max(parts.count - 1, 0)
var result = ""
result.reserveCapacity(capacity)
var iter = parts.makeIterator()
if let first = iter.next() {
    result.append(contentsOf: first)
    while let next = iter.next() {
        result.append("\\")
        result.append(contentsOf: next)
    }
}
print("normalized (updated): " + (result.isEmpty ? "." : result))
```

Here is a standalone script to demonstrate the behavioral changes before and after this patch. I'd treat this as a **huge** and **urgent** improvement. The current one is so buggy that it needs to be replaced ASAP on all the active branches.

```cmd
C:\Users\stevapple>normalize.exe                                 
user input: ..\swift-tools-support-core
normalized (current): swift-tools-support-core
normalized (updated): ..\swift-tools-support-core

C:\Users\stevapple>normalize.exe ..\hello\hello\..\my-package
user input: ..\hello\hello\..\my-package
normalized (current): hello\my-package
normalized (updated): ..\hello\my-package

C:\Users\stevapple>normalize.exe hello\..\..\hello2\..\my-package   
user input: hello\..\..\hello2\..\my-package
normalized (current): \my-package
normalized (updated): ..\my-package

C:\Users\stevapple>normalize.exe .\hello\..\.\hello
user input: .\hello\..\.\hello
normalized (current): \hello
normalized (updated): hello
```

CC: @compnerd @abertelrud

compnerd:
> Why I regard it as critical — Local debugging relies largely on relative paths, and for now we’re not even able to set the path of a dependency package to `../package`. This is unacceptable for user experience and will trouble toolchain development on Windows too.

I strongly disagree with this statement and see this as purely hyperbole, which really does not add much to the discussion.  I definitely am using SPM actively both for local and CI purposes, it definitely works fine for local debugging.

> Here is a standalone script to demonstrate the behavioral changes on Windows before and after this patch. I'd treat this as a huge and urgent improvement. The current one is so buggy that it needs to be replaced ASAP on all the active branches.

That doesn't change the fact that the handling does not work for a variety of paths on Windows.  This is a very large change that fixes a very particular case (which would require you to go outside of the source tree, which is really not how SPM normally operates).  Thus, this is trying to fix a corner case for a corner case.  I think that the efforts are better spent on porting to System.

Alternatively, I would be fine with temporarily unifying the paths with Foundation.

stevapple:
Succeeded by #221

elsh:
Closing as it's superseded by https://github.com/apple/swift-tools-support-core/pull/219

PR #199, opened 2021-03-18
restore disabled FileSystemTests

tomerd:
motivation: better test coverage

changes:
* refactor the common part of testInMemoryFileSystemFileLock, testLocalFileSystemFileLock, testRerootedFileSystemViewFileLock
* resove the potential reader race by making sure file have content prior to starting the threads
* enable the above disabled tests in FileSystemTests

rdar://71560894

### DIFF
--- a/Tests/TSCBasicTests/FileSystemTests.swift
+++ b/Tests/TSCBasicTests/FileSystemTests.swift
@@ -749,9 +749,6 @@ class FileSystemTests: XCTestCase {
     }
 
     func testInMemoryFileSystemFileLock() throws {
-        // Disabled until rdar://71560894 is fixed.
-        try XCTSkipIf(true)
-        
         let fs = InMemoryFileSystem()
         let path = AbsolutePath("/")
         try fs.createDirectory(path)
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci please test

Tests/TSCBasicTests/FileSystemTests.swift
```diff
@@ -884,22 +783,31 @@ class FileSystemTests: XCTestCase {
         let fileB = path.appending(component: "fileB")
         let lockFile = path.appending(component: "lockfile")
 
+        try _testFileSystemFileLock(fileSystem: fs, fileA: fileA, fileB: fileB, lockFile: lockFile)
+    }
+
+    private func _testFileSystemFileLock(fileSystem fs: FileSystem, fileA: AbsolutePath, fileB: AbsolutePath, lockFile: AbsolutePath) throws {
+
+        // write initial value, since reader may start before writers and files would not exist
+        try fs.writeFileContents(fileA, bytes: [0])
```
benlangmuir, line 792: This is writing a null byte, so the code below that tries to read an integer as a string from the file will still fail.  I guess it's "okay" in the sense that it will then hit the `?? 0`, but it seems kind of indirect and hard to follow, especially since it's easy to misread writing 0 as being equivalent to writing "0".  Is there a reason not to write "0" to the file instead of "\0" and then simplify the logic below to force the optional Int?

Tests/TSCBasicTests/FileSystemTests.swift
```diff
@@ -884,22 +783,31 @@ class FileSystemTests: XCTestCase {
...
                     // Sum and write back to file.
-                    try! fs.writeFileContents(fileA, bytes: ByteString(encodingAsUTF8: String(valueA + 1)))
+                    try fs.writeFileContents(fileA, bytes: ByteString(encodingAsUTF8: String(valueA + 1)))
 
                     Thread.yield()
 
-                    // Get thr current contents of the file if any.
+                    // Get the current contents of the file if any.
                     let valueB: Int
                     if fs.exists(fileB) {
```
benlangmuir, line 812: Always true now

Tests/TSCBasicTests/FileSystemTests.swift
tomerd: good point, I can revise

tomerd:
@swift-ci please test

neonichu (approved):

PR #205, opened 2021-04-07
fix flaky test that can hang in CI

tomerd:
motivation: stable CI

background:
* testSignal can hang *sometimes* when the child processes it creates do not recieve the expeted signal.
* this has been observed when waitForFile timed-out and the test exited before issuing the signal.
* this behavior is unexpected and points to potential issue with xctest, but this fix attempts to workaround the underlying issue by making sure the sub-processes are forced-killed by the test itself

changes:
* regsiter the sub-processes in a process-set and force temrinate them at the end of the test

rdar://74356445


tomerd:
@swift-ci please test

abertelrud (approved):
This looks good to me.

neonichu (approved):

PR #206, opened 2021-04-07
increase waitForFile test utility timeout

tomerd:
motivation: on busy systems (such as CI) this could take a while

changte: set timeout to 30s which should be plenty enough


tomerd:
@swift-ci please test

abertelrud (approved):

PR #209, opened 2021-04-14
improve process state management

tomerd:
motivation: reports on hang process (mostly tests)

changes:
* refactor Process to use a state machine to track the process execution state
* replace use of DispatchQueue with Locks to protect state

rdar://76087764

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -181,7 +181,8 @@ public final class Process: ObjectIdentifierProtocol {
     // process execution mutable state
     private enum State {
         case idle
-        case readingOutput(stdout: Thread, stderr: Thread?)
+        case readingOutputThread(stdout: Thread, stderr: Thread?)
+        case readingOutputPipe(sync: DispatchGroup)
         case outputReady(stdout: Result<[UInt8], Swift.Error>, stderr: Result<[UInt8], Swift.Error>)
         case complete(ProcessResult)
     }
@@ -232,7 +233,7 @@ public final class Process: ObjectIdentifierProtocol {
     private let stateLock = Lock()
 
     /// The result of the process execution. Available after process is terminated.
-    /// This will block while the process is running, as such equivalent to `waitUntilExit`
+    /// This will block while the process is awaiting result
     @available(*, deprecated, message: "use waitUntilExit instead")
     public var result: ProcessResult? {
         return self.stateLock.withLock {
@@ -403,42 +404,64 @@ public final class Process: ObjectIdentifierProtocol {
         let stdinPipe = Pipe()
         _process?.standardInput = stdinPipe
 
+        let group = DispatchGroup()
+
+        var stdout: [UInt8] = []
+        let stdoutLock = Lock()
+
+        var stderr: [UInt8] = []
+        let stderrLock = Lock()
+
         if outputRedirection.redirectsOutput {
             let stdoutPipe = Pipe()
             let stderrPipe = Pipe()
 
-            var pending: [UInt8]?
-            let pendingLock = Lock()
-
+            group.enter()
             stdoutPipe.fileHandleForReading.readabilityHandler = { (fh : FileHandle) -> Void in
-                let contents = fh.readDataToEndOfFile()
-                self.outputRedirection.outputClosures?.stdoutClosure([UInt8](contents))
-                pendingLock.withLock {
-                    if let stderr = pending {
-                        self.stateLock.withLock {
-                            self?.state = .outputReady(stdout: .success(contents), stderr: .success(stderr))
-                        }
-                    } else {
-                        pending = contents
+                let data = fh.availableData
+                if (data.count == 0) {
+                    stdoutPipe.fileHandleForReading.readabilityHandler = nil
+                    group.leave()
+                } else {
+                    let contents = data.withUnsafeBytes { Array<UInt8>($0) }
+                    self.outputRedirection.outputClosures?.stdoutClosure(contents)
+                    stdoutLock.withLock {
+                        stdout += contents
                     }
                 }
             }
+
+            group.enter()
             stderrPipe.fileHandleForReading.readabilityHandler = { (fh : FileHandle) -> Void in
-                let contents = fh.readDataToEndOfFile()
-                self.outputRedirection.outputClosures?.stderrClosure([UInt8](contents))
-                pendingLock.withLock {
-                    if let stdout = pending {
-                        self.stateLock.withLock {
-                            self?.state = .outputReady(stdout: .success(stdout), stderr: .success(contents))
-                        }
-                    } else {
-                        pending = contents
+                let data = fh.availableData
+                if (data.count == 0) {
+                    stderrPipe.fileHandleForReading.readabilityHandler = nil
+                    group.leave()
+                } else {
+                    let contents = data.withUnsafeBytes { Array<UInt8>($0) }
+                    self.outputRedirection.outputClosures?.stderrClosure(contents)
+                    stderrLock.withLock {
+                        stderr += contents
                     }
                 }
             }
 
             _process?.standardOutput = stdoutPipe
             _process?.standardError = stderrPipe
+        }
+
+        // first set state then start reading threads
+        let sync = DispatchGroup()
+        sync.enter()
+        self.stateLock.withLock {
+            self.state = .readingOutputPipe(sync: sync)
+        }
+
+        group.notify(queue: .global()) {
+            self.stateLock.withLock {
+                self.state = .outputReady(stdout: .success(stdout), stderr: .success(stderr))
+            }
+            sync.leave()
         }
 
         try _process?.run()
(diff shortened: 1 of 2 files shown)


tomerd:
brings back #203

tomerd:
@compnerd I tired to code up the windows part but have no way to test, could you ptal and help get over the finish line

tomerd:
@swift-ci please test

neonichu (approved):

Sources/TSCBasic/Process.swift
```diff
@@ -219,36 +227,36 @@ public final class Process: ObjectIdentifierProtocol {
...
-    /// If the subprocess has launched.
-    /// Note: This property is not protected by the serial queue because it is only mutated in `launch()`, which will be
-    /// called only once.
-    public private(set) var launched = false
+    // process execution mutable state
+    private var state: State = .idle
+    private let stateLock = Lock()
 
     /// The result of the process execution. Available after process is terminated.
+    /// This will block while the process is running, as such equivalent to `waitUntilExit`
```
abertelrud, line 235: Based on the locking here it doesn't look as if it will block while the process is running, but rather return nil.  This is because I don't think `stateLock` is held all the time while the process is running (it's unlocked again on line 681).  So this comment is probably not accurate.

abertelrud: (it does make sense that `stateLock` is only used to protect `state`, BTW, and isn't held for the duration of the process — if this property should block, could it instead just call `waitUntilExit`).

tomerd, line 235:
```suggestion
    /// This will block while the process is awaiting result
```

tomerd:
@swift-ci please test

Sources/TSCBasic/Process.swift
tomerd: it blocks while awaiting results (specifically `waitpid`), reworded accordingly

compnerd:
This definitely doesn't seem to build on Windows:

```
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:419:33: error: cannot use optional chaining on non-optional value of type 'Process'
                            self?.state = .outputReady(stdout: .success(contents), stderr: .success(stderr))
                            ~~~~^

S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:419:73: error: cannot convert value of type 'Data' to expected argument type '[UInt8]'
                            self?.state = .outputReady(stdout: .success(contents), stderr: .success(stderr))
                                                                        ^
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:422:35: error: cannot assign value of type 'Data' to type '[UInt8]'
                        pending = contents
                                  ^~~~~~~~
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:432:33: error: cannot use optional chaining on non-optional value of type 'Process'
                            self?.state = .outputReady(stdout: .success(stdout), stderr: .success(contents))
                            ~~~~^

S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:432:99: error: cannot convert value of type 'Data' to expected argument type '[UInt8]'
                            self?.state = .outputReady(stdout: .success(stdout), stderr: .success(contents))
                                                                                                  ^
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:435:35: error: cannot assign value of type 'Data' to type '[UInt8]'
                        pending = contents
                                  ^~~~~~~~
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:660:16: error: value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_iobuf>') has no member 'thread'
        stdout.thread?.join()
        ~~~~~~ ^~~~~~
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:661:16: error: value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_iobuf>') has no member 'thread'
        stderr.thread?.join()
        ~~~~~~ ^~~~~~
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:667:28: error: value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_iobuf>') has no member 'result'
            output: stdout.result,
                    ~~~~~~ ^~~~~~
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:668:34: error: value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_iobuf>') has no member 'result'
            stderrOutput: stderr.result
                          ~~~~~~ ^~~~~~
```

Sources/TSCBasic/Process.swift
```diff
@@ -396,27 +406,44 @@ public final class Process: ObjectIdentifierProtocol {
...
             }
             stderrPipe.fileHandleForReading.readabilityHandler = { (fh : FileHandle) -> Void in
                 let contents = fh.readDataToEndOfFile()
                 self.outputRedirection.outputClosures?.stderrClosure([UInt8](contents))
-                if case .success(let data) = self.stderr.result {
-                    self.stderr.result = .success(data + contents)
+                pendingLock.withLock {
+                    if let stdout = pending {
+                        self.stateLock.withLock {
+                            self?.state = .outputReady(stdout: .success(stdout), stderr: .success(contents))
```
compnerd, line 432:
```suggestion
                            self.state = .outputReady(stdout: .success(stdout), stderr: .success(contents.withUnsafeBytes { Array<UInt8>($0) }))
```

Sources/TSCBasic/Process.swift
```diff
@@ -396,27 +406,44 @@ public final class Process: ObjectIdentifierProtocol {
...
+                        }
+                    } else {
+                        pending = contents
```
compnerd, line 435:
```suggestion
                        pending = contents.withUnsafeBytes { Array<UInt8>($0) }
```

Sources/TSCBasic/Process.swift
```diff
@@ -396,27 +406,44 @@ public final class Process: ObjectIdentifierProtocol {
...
                 self.outputRedirection.outputClosures?.stdoutClosure([UInt8](contents))
-                if case .success(let data) = self.stdout.result {
-                    self.stdout.result = .success(data + contents)
+                pendingLock.withLock {
+                    if let stderr = pending {
+                        self.stateLock.withLock {
+                            self?.state = .outputReady(stdout: .success(contents), stderr: .success(stderr))
+                        }
+                    } else {
+                        pending = contents
```
compnerd, line 422:
```suggestion
                        pending = contents.withUnsafeBytes { Array<UInt8>($0) }
```

Sources/TSCBasic/Process.swift
```diff
@@ -396,27 +406,44 @@ public final class Process: ObjectIdentifierProtocol {
...
+
             stdoutPipe.fileHandleForReading.readabilityHandler = { (fh : FileHandle) -> Void in
                 let contents = fh.readDataToEndOfFile()
                 self.outputRedirection.outputClosures?.stdoutClosure([UInt8](contents))
-                if case .success(let data) = self.stdout.result {
-                    self.stdout.result = .success(data + contents)
+                pendingLock.withLock {
+                    if let stderr = pending {
+                        self.stateLock.withLock {
+                            self?.state = .outputReady(stdout: .success(contents), stderr: .success(stderr))
```
compnerd, line 419:
```suggestion
                            self.state = .outputReady(stdout: .success(contents.withUnsafeBytes { Array<UInt8>($0) }), stderr: .success(stderr))
```

compnerd (commented):
The `waitUntilExit` is also something that will need to be updated for your structural changes.

Sources/TSCBasic/Process.swift
abertelrud: Based on the current implementation, the semantics seem to be that it returns nil if called in `readingOutput` state (due to not holding the state lock during the `Thread.join()` calls) but blocks if called in `outputReady` state (due to holding the state lock during `waitpid()`).  So it depends on whether all the output from the process has been read yet.  Since the `result` property is deprecated and nothing seems to rely on it, I don't think that's a problem, but wanted to check if that matches the intent.

tomerd: short of removing this method, this is the only safe option I can think of. do you have other ideas we should explore?

tomerd:
@swift-ci please test

compnerd:
This builds on Windows, but SPM no longer runs.

tomerd:
@compnerd is this ready now?

tomerd:
@swift-ci please test

tomerd:
@swift-ci please test linux

tomerd:
@compnerd can you verify this is good to go?

compnerd:
@tomerd yeah, given our work on this, I think that this is okay to merge.  I've been running with it locally for ~1 week, and I think that the only "downside" to it is that it is a slight bit slower on the startup.

PR #213, opened 2021-05-07
Deserialize diagnostics with no location correctly

owenv:
When a diagnostic doesn't have a source location, no filename record will be emitted. This PR makes the filename of `SourceLocation` optional so these can be deserialized successfully.


owenv:
@swift-ci please test

Sources/TSCUtility/SerializedDiagnostics.swift
```diff
@@ -185,7 +185,7 @@ extension SerializedDiagnostics {
 
   public struct SourceLocation: Equatable {
     /// The filename associated with the diagnostic.
-    public var filename: String
+    public var filename: String?
```
abertelrud, line 188: What do the line and column mean (semantically) in the absence of a filename?  If the .dia still has those fields without the filename then this does seem like the right change, but what should an IDE (for example) do with such a source location?

abertelrud: Or should this be modeled so that it's the `SourceLocation` itself that is optional?  (even if that's indicated by just a `nil` filename field in the file).

abertelrud (approved):
Looks good to me, but had one question about whether it's the filename or the whole source location that should be optional, semantically.

BTW, I'm really hoping to soon be able to start making use of this native ability to deserialize diagnostics files in SwiftPM.  There are places where it's looking at the stderr of the compiler output and it's going to be great to start cleaning those up!

Sources/TSCUtility/SerializedDiagnostics.swift
owenv: I updated this to just make the entire SourceLocation optional. Currently, Swift uses the absence of a filename to indicate either a missing or invalid location, and it's always paired with a line/column of 0. I'm fairly certain diagnostics formatted from Clang will look the same, but even if it encodes additional info there's really nothing an IDE can do to present that in a useful way.

owenv:
@swift-ci please test

owenv:
@swift-ci please smoke test

Sources/TSCUtility/SerializedDiagnostics.swift
abertelrud: That makes a lot of sense to me.  Thanks!

abertelrud:
@swift-ci please test

owenv:
@swift-ci please test Linux

owenv:
` error: module file '/home/buildnode/.cache/clang/ModuleCache/315MAVBT9WZMX/SwiftShims-339366MV89AHO.pcm' is out of date and needs to be rebuilt: signature mismatch`\

@swift-ci please test Linux

tomerd:
@owenv do you want to cherry pick this into release/5.5 too?

owenv:
@tomerd thanks for merging! None of the higher-level tools use this code on the 5.5 branches, so I don't think it's worth cherry-picking.

PR #217, opened 2021-05-26
TSCBasic: unify subprocess handling

compnerd:
This unifies the platforms to Foundation for the process handling.  This
simplifies the logic and gives a single codepath which can possibly
serve as an alternate path to #209.


compnerd:
@swift-ci please test

PR #219, opened 2021-06-06
[DNM] Refactor Path APIs with SwiftSystem

stevapple:
_This patch contains API changes. Minor version bump required._

Refactor Path APIs with SwiftSystem, and meanwhile align some APIs’ behavior. This includes:
- No longer allow blank component `""`. This was in the doc but not actually take into practice.
- No longer return root component in `AbsolutePath.components`.  Add `root` for all `Path` instances.
- Creating `RelativePath` against another `AbsolutePath` may throw now, because `AbsolutePath`s could have different roots on Windows.
- Deprecate `AbsolutePath.root` for better stability and clearer usage.

This PR needs parallel changes in all the dependent packages:
- swift-llbuild
- swift-package-manager
- sourcekit-lsp
- swift-driver
- swift-llbuild2
- swiftpm-on-llbuild2
- swift-stress-tester
- swift-tools-support-async

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,7 +28,9 @@ let package = Package(
             name: "TSCTestSupport",
             targets: ["TSCTestSupport"]),
     ],
-    dependencies: [],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-system.git", .upToNextMinor(from: "0.0.1"))
+    ],
     targets: [
         
         // MARK: Tools support core targets
@@ -44,7 +46,8 @@ let package = Package(
         .target(
             /** TSCBasic support library */
             name: "TSCBasic",
-            dependencies: ["TSCLibc", "TSCclibc"]),
+            dependencies: ["TSCLibc", "TSCclibc",
+                           .product(name: "SystemPackage", package: "swift-system")]),
         .target(
             /** Abstractions for common operations, should migrate to TSCBasic */
             name: "TSCUtility",
--- a/Package.swift
+++ b/Package.swift
@@ -77,16 +77,3 @@ let package = Package(
             dependencies: ["TSCUtility", "TSCTestSupport"]),
     ]
 )
-
-// FIXME: conditionalise these flags since SwiftPM 5.3 and earlier will crash
-// for platforms they don't know about.
-#if os(Windows)
-  if let TSCBasic = package.targets.first(where: { $0.name == "TSCBasic" }) {
-    TSCBasic.cxxSettings = [
-      .define("_CRT_SECURE_NO_WARNINGS", .when(platforms: [.windows])),
-    ]
-    TSCBasic.linkerSettings = [
-      .linkedLibrary("Pathcch", .when(platforms: [.windows])),
-    ]
-  }
-#endif
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -1,25 +1,172 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+ Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
-#if os(Windows)
-import Foundation
-import WinSDK
-#endif
-
-#if os(Windows)
-private typealias PathImpl = UNIXPath
-#else
-private typealias PathImpl = UNIXPath
-#endif
+import SystemPackage
 
 import protocol Foundation.CustomNSError
 import var Foundation.NSLocalizedDescriptionKey
+
+public protocol Path: Hashable, Codable, CustomStringConvertible {
+    /// Underlying type, based on SwiftSystem.
+    var filepath: FilePath { get }
+    
+    /// Public initializer from FilePath.
+    init(_ filepath: FilePath)
+    
+    /// Public initializer from String.
+    init(_ string: String)
+    
+    /// Convenience initializer that verifies that the path lexically.
+    init(validating path: String) throws
+    
+    /// Normalized string representation (the normalization rules are described
+    /// in the documentation of the initializer).  This string is never empty.
+    var pathString: String { get }
+
+    /// The root of a path.
+    var root: String? { get }
+
+    /// Directory component.  An absolute path always has a non-empty directory
+    /// component (the directory component of the root path is the root itself).
+    var dirname: String { get }
+
+    /// Last path component (including the suffix, if any).
+    var basename: String { get }
+    
+    /// Returns the basename without the extension.
+    var basenameWithoutExt: String { get }
+    
+    /// Extension of the give path's basename. This follow same rules as
+    /// suffix except that it doesn't include leading `.` character.
+    var `extension`: String? { get }
+    
+    /// Suffix (including leading `.` character) if any.  Note that a basename
+    /// that starts with a `.` character is not considered a suffix, nor is a
+    /// trailing `.` character.
+    var suffix: String? { get }
+    
+    /// True if the path is a root directory.
+    var isRoot: Bool { get }
+    
+    /// Returns the path with an additional literal component appended.
+    ///
+    /// This method accepts pseudo-path like '.' or '..', but should not contain "/".
+    func appending(component: String) -> Self
+    
+    /// Returns the relative path with additional literal components appended.
+    ///
+    /// This method should only be used in cases where the input is guaranteed
+    /// to be a valid path component (i.e., it cannot be empty, contain a path
+    /// separator, or be a pseudo-path like '.' or '..').
+    func appending(components names: [String]) -> Self
+    func appending(components names: String...) -> Self
+    
+    /// Returns an array of strings that make up the path components of the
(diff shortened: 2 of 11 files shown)


stevapple:
Please point out if we can minimize the impact while keeping correct behavior.

Also CC @compnerd @milseman

Package.swift
```diff
@@ -28,7 +28,9 @@ let package = Package(
             name: "TSCTestSupport",
             targets: ["TSCTestSupport"]),
     ],
-    dependencies: [],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-system.git", .upToNextMinor(from: "0.0.1"))
```
milseman, line 32: Should this be `0.0.1` or `0.0.2`?

Sources/TSCBasic/Path.swift
```diff
@@ -1,26 +1,178 @@
...
+        return filepath.isRoot
+    }
+
+    public func appending(component: String) -> Self {
+        return Self(filepath.appending(
+                                FilePath.Component(stringLiteral: component)))
+    }
+
+    public func appending(components names: [String]) -> Self {
+        let components = names.map(FilePath.Component.init)
```
milseman, line 134: You can also consider throwing in a `.lazy`.

Sources/TSCBasic/Path.swift
```diff
@@ -179,24 +255,50 @@ public struct AbsolutePath: Hashable {
     /// NOTE: We may want to consider adding operators such as `+` for appending
     ///       a path component.
 
-    /// NOTE: We will want to add a method to return the lowest common ancestor
-    ///       path.
-
-    /// Root directory (whose string representation is just a path separator).
-    public static let root = AbsolutePath(PathImpl.root)
+    /// Returns the lowest common ancestor path.
+    public func lowestCommonAncestor(with path: AbsolutePath) -> AbsolutePath? {
```
milseman, line 259: Common ancestor queries is also a good candidate for FilePath. We had operations like `lexicallyRelative(to:)` in earlier pitches.

Sources/TSCBasic/Path.swift
```diff
@@ -865,53 +387,37 @@ extension AbsolutePath {
...
-        // FIXME: The is needs to be optimized to avoid unncessary copying.
-        let pathComps = self.components
-        let baseComps = base.components
-
-        // It's common for the base to be an ancestor, so try that first.
-        if pathComps.starts(with: baseComps) {
-            // Special case, which is a plain path without `..` components.  It
-            // might be an empty path (when self and the base are equal).
-            let relComps = pathComps.dropFirst(baseComps.count)
+    public func relative(to base: AbsolutePath) throws -> RelativePath {
```
milseman, line 390: Another candidate for `FilePath`.

Sources/TSCBasic/PathShims.swift
```diff
@@ -168,18 +168,20 @@ extension AbsolutePath {
...
             return self.pathString
         }
-        // FIXME: Instead of string prefix comparison we should add a proper API
-        // to AbsolutePath to determine ancestry.
-        if self == dir {
-            return "."
-        } else if self.pathString.hasPrefix(dir.pathString + "/") {
-            return "./" + self.relative(to: dir).pathString
+        if let first = rel.components.first,
+            first != ".." {
```
milseman, line 177: What is the situation you're checking here? `..`s should be removed as part of lexical normalization except at the start of relative paths.

milseman (approved):
This LGTM. I'm not a domain expert on SPM or this use case. @compnerd what do you think about this approach?

(Sorry for the delay. I had a 90% completed review but got distracted before I could finish, and forgot about it)

Package.swift
stevapple: It should be `0.0.3` given that the first two releases actually not compiles (`0.0.2` only compiles with release config).

`upToNextMinor()` indicates the latest `0.0.x`, once `0.0.3` is released it should resolve.

Sources/TSCBasic/PathShims.swift
stevapple: I just copied the old behavior here. The old code shows that the returned string starts with either `..` or `.`.

Sources/TSCBasic/Path.swift
```diff
@@ -1,26 +1,178 @@
...
+    /// Underlying type, based on SwiftSystem.
+    var filepath: FilePath { get }
+
+    /// Public initializer from FilePath.
+    init(_ filepath: FilePath)
+
+    /// Public initializer from String.
+    init(_ string: String)
+
+    /// Convenience initializer that verifies that the path lexically.
```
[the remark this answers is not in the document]
stevapple: The initializer will throw if the argument isn’t a valid relative/absolute path.

Sources/TSCBasic/Path.swift
```diff
@@ -179,24 +255,50 @@ public struct AbsolutePath: Hashable {
...
+    @available(*, deprecated, message: "root is not a static value, use the instance property instead")
+    public static var root: AbsolutePath {
+        if let rootPath = localFileSystem.currentWorkingDirectory?.root {
+            return AbsolutePath(rootPath)
+        } else {
+#if !os(Windows)
+            return AbsolutePath("/")
+#else
+            if let drive = ProcessEnv.vars["SystemDrive"] ?? ProcessEnv.vars["HomeDrive"] {
+                return AbsolutePath(drive + "\\")
```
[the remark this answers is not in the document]
stevapple: `%SystemDrive%` appears like `C:`, and `C:\` is the file path of its root.

Sources/TSCBasic/Path.swift
```diff
@@ -1,26 +1,178 @@
...
+    func appending(components names: [String]) -> Self
+    func appending(components names: String...) -> Self
+
+    /// Returns an array of strings that make up the path components of the
+    /// path.  This is the same sequence of strings as the basenames of each
+    /// successive path component.  An empty path has a single path
+    /// component: the `.` string.
+    ///
+    /// NOTE: Path components no longer include the root.  Use `root` instead.
+    var components: [String] { get }
```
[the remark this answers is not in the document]
stevapple: > What's the main reason for this protocol in SPM?
These are just common APIs of `RelativePath` and `AbsolutePath`.
> How many of them should be dropped or superseded by `FilePath`?
I don’t think I’m capable enough of answering this question. If we’d like to drop any of them, maybe we’d get other TSC specialists in here.

Sources/TSCBasic/Path.swift
```diff
@@ -179,24 +255,50 @@ public struct AbsolutePath: Hashable {
...
-    public var components: [String] {
-        return _impl.components
+    public static func withPOSIX(path: String) -> AbsolutePath {
+#if os(Windows)
+        var filepath = FilePath(path)
+        precondition(filepath.root != nil)
+        if !filepath.isAbsolute {
+            filepath.root = root.filepath.root
+        }
+        return AbsolutePath(filepath)
```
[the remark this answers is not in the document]
stevapple: `public static func withPOSIX(path: String) -> AbsolutePath` is intended for using POSIX-like file systems that have a single root on Windows. Therefore, the `path` should always begin with `/`, then implicitly transformed to `\`, which is a valid root for Windows.

Correspondingly, if `path` doesn’t begin with `/`, `AbsolutePath(path)` will have precondition failures on macOS and Linux.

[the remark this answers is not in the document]
stevapple: It would be better if we have `PosixFilePath` and `WindowsFilePath`, with conditional `typealias FilePath` to the default one. POSIX file paths are useful sometimes on Windows.

tomerd:
@stevapple what is the status of this work? is this something you want and can bring over the finish line?

stevapple:
@tomerd I’m afraid not yet. Here’s a simple checklist:
- We would like to keep using POSIX path style for in-memory file systems, and POSIX `FilePath` is not supported on Windows yet. We’d better get @milseman in to see if it’s possible to add support for that. This would help to remove the `.posix()` workaround, which largely reduces API breakage.
- Because this is an API breaking change, we also need to update and dependent projects against it. Such work needs to be done after the API is fixed.

@milseman also suggests to move some implementation details to `swift-system`, but this could be addressed in another PR.

tomerd:

stevapple:
> @stevapple @compnerd we now have swift-system integrated into the main branch of TSC and can start making use of it / reduce code to help support windows paths etc better

IMO `FilePath` is too platform-specific to implement the general interface of TSC. For example, `InMemoryFileSystem` can only represent a simple FS, which falls to a rather small part of the Windows path system. By using `FilePath`, we need to take extra care about these unsupported paths.

My suggested migration path is to directly rewriting TSC's clients with `FilePath`, until we can finally deprecate the whole path submodule in TSC.

I'm also working on a high-level and cross-platform path package (something like karwa/swift-url), but I think that help little to TSC.

stevapple:
@tomerd @compnerd @milseman 

While it's still undecided whether we're able to split `POSIXPath` out of `FilePath`, it's now possible for us to start the migration.

I introduced a convention that will automatically rebase a `/`-based path string to the system root, which enables smoother transition for non-Windows file systems.

The only breaking change now is that `AbsolutePath.relative(to:)` throws. It has a non-throwing counterpart `AbsolutePath.init(_:_:)` so I think it'll be easier for other tools to adapt. It rarely throws in production IMO.

stevapple:
For anyone who may be concerned, you can start to review the code now.

I would investigate the impact upon TSC clients and open corresponding PR to test it with CI.

PR #221, opened 2021-06-06
[Experimental][DNM] Use SwiftSystem for file operations

stevapple:
This PR intends to use `SwiftSystem` for file operations in `TSCBasic`, which are currently based on `TSClibc`.
It is based on #219 .

This will:
- Simplify the implementation;
- Make the code a lot more “Swifty”;
- Potentially improve the performance.

Discussion: Do we really need a buffer for I/O?

TBD: (mostly `SwiftSystem` side)
- Pipe support;
- Better `FileDescriptor` to `FILE` bridge.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "swift-system",
+        "repositoryURL": "https://github.com/apple/swift-system.git",
+        "state": {
+          "branch": null,
+          "revision": "2bc160bfe34d843ae5ff47168080add24dfd7eac",
+          "version": "0.0.2"
+        }
+      }
+    ]
+  },
+  "version": 1
+}
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -8,9 +8,9 @@
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
-import TSCLibc
 import Foundation
 import Dispatch
+import SystemPackage
 
 public struct FileSystemError: Error, Equatable {
     public enum Kind: Equatable {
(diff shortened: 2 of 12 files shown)


Package.resolved
```diff
@@ -0,0 +1,16 @@
...
+        "state": {
+          "branch": null,
+          "revision": "2bc160bfe34d843ae5ff47168080add24dfd7eac",
+          "version": "0.0.2"
+        }
+      }
+    ]
+  },
+  "version": 1
+}
```
tomerd, line 16: please remove this file

Sources/TSCBasic/FileSystem.swift
```diff
@@ -86,23 +86,34 @@ extension FileSystemError: CustomNSError {
...
-        case TSCLibc.EISDIR:
-            self.init(.isDirectory, path)
-        case TSCLibc.ENOENT:
-            self.init(.noEntry, path)
-        case TSCLibc.ENOTDIR:
-            self.init(.notDirectory, path)
-        default:
-            self.init(.unknownOSError, path)
+        case .permissionDenied:         return .invalidAccess
+        case .isDirectory:              return.isDirectory
```
tomerd, line 93: missing space?

Sources/TSCBasic/Path.swift
```diff
@@ -865,53 +366,37 @@ extension AbsolutePath {
...
+        /// TODO: DOS relative path may change the root.
+        if root != base.root {
+            throw PathValidationError.differentRoot(pathString, base.pathString)
         }
-        assert(base.appending(result) == self)
-        return result
+#endif
+        filepath.root = base.filepath.root
+
+        let commonAncestor = AbsolutePath(filepath).lowestCommonAncestor(with: base)!
```
tomerd, line 380: guard + precondition instead of bang?

Sources/TSCBasic/Path.swift
```diff
@@ -1,26 +1,178 @@
...
-private typealias PathImpl = UNIXPath
-#else
-private typealias PathImpl = UNIXPath
-#endif
+import SystemPackage
 
 import protocol Foundation.CustomNSError
 import var Foundation.NSLocalizedDescriptionKey
 
+public protocol Path: Hashable, Codable, CustomStringConvertible {
```
tomerd, line 15: wonder if we should pick a different name for these types, as ideally we deprecate them all over time

Sources/TSCBasic/Path.swift
stevapple: I don’t think callers should guard such condition🤔 Breaking is quite reasonable and necessary here.

We’d better discuss in #219 since this API change is introduced there.

Sources/TSCBasic/Path.swift
stevapple: Use another name in #219

tomerd:

PR #235, opened 2021-06-29
Default to macOS 10.10 minimum deployment target again but allow it to be overridden

abertelrud:
This reverts the main part of the https://github.com/apple/swift-tools-support-core/pull/225 change.

This reverts commit 5e968d4b3bb92e4cd39efdaac56f0a05de122f1e since it caused breakage, but keeps the iOS 13 minimum deployment target.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,10 +14,17 @@
 import PackageDescription
 import class Foundation.ProcessInfo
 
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please test

PR #236, opened 2021-06-29
Improve Deserialization Performance for Bitstream Files

CodaFi:
Switch to a non-owning intermediate representation to allow the caller to perform the necessary copies when it reinterprets the raw data we provide to them in the visitor. This meant removing the remaining entry points that could possibly allow unowned data to escape - so the initializers on `Bitstream` that ran the `CollectingVisitor` have been struck - as well as the existing deprecated APIs here that we no longer need.

The combo of all of the changes here improves the performance of dependency graph deserialization in the Swift driver by 3-5x which translates to a 3-5 second wall time improvement for large projects (> 1000 files).

### DIFF
--- a/Sources/TSCUtility/BitstreamReader.swift
+++ b/Sources/TSCUtility/BitstreamReader.swift
@@ -148,6 +148,10 @@ private struct BitstreamReader {
     }
   }
 
+  /// Computes a non-owning view of a `BitcodeElement.Record` that is valid for
+  /// the lifetime of the call to `body`.
+  ///
+  /// - Warning: If this function throws, the `body` block will not be called.
   mutating func withAbbreviatedRecord(
     _ abbrev: Bitstream.Abbreviation,
     body: (BitcodeElement.Record) throws -> Void
--- a/Sources/TSCUtility/BitstreamReader.swift
+++ b/Sources/TSCUtility/BitstreamReader.swift
@@ -176,13 +176,29 @@ private struct BitstreamReader {
       switch lastOperand {
       case .array(let element):
         let length = try cursor.readVBR(6)
-        var elements = [UInt64]()
-        for _ in 0..<length {
-          elements.append(try readSingleAbbreviatedRecordOperand(element))
-        }
         if case .char6 = element {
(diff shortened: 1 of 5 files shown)


Sources/TSCUtility/BitstreamReader.swift
```diff
@@ -222,26 +180,42 @@ private struct BitstreamReader {
...
-        payload = .blob(try Data(cursor.read(bytes: length)))
+        payload = .blob(try cursor.read(bytes: length))
         try cursor.advance(toBitAlignment: 32)
       default:
         fatalError()
       }
     }
 
-    return .init(id: code, fields: fields, payload: payload)
+    return try body(.init(id: code, fields: UnsafeBufferPointer(fields), payload: payload))
```
cltnschlosser, line 218: How does this work? If fields is deallocated in the defer block?

Sources/TSCUtility/BitstreamWriter.swift
```diff
@@ -367,6 +368,7 @@ extension BitstreamWriter {
         }
 
         public mutating func append(_ string: String) {
+            self.values.reserveCapacity(self.values.capacity + string.count)
```
cltnschlosser, line 371: string.utf8.count

Sources/TSCUtility/BitstreamReader.swift
CodaFi: `defer` blocks run at scope exit. The order of operations is 
1) Construct fields
2) Invoke continuation
3) Execute defers

cltnschlosser: Oh, duh. Thanks.

neonichu (approved):

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

abertelrud (approved):

CodaFi:
@swift-ci test

CodaFi:
apple/swift-driver#734

@swift-ci test

CodaFi:
⛵

PR #239, opened 2021-07-01
TSCUtility: implement FSWatch for Windows

compnerd:
Rather than simply take the no-op watcher which does not watch the file
system, implement the FS Watcher using `ReadDirectoryChangesW`.  This
should allow SourceKit-LSP to be aware of changes to the
package.resolved file.


compnerd:
@swift-ci please test

tomerd (approved):

PR #241, opened 2021-07-02
Add conditionals to unbreak building ToolsSupportCore for iOS

abertelrud:
One availability conditional wasn't properly including the iOS platform, and another conditional caused an "implementation required" error.  This restores the ability to build on iOS after #228 and #236.

### DIFF
--- a/Sources/TSCUtility/BitstreamReader.swift
+++ b/Sources/TSCUtility/BitstreamReader.swift
@@ -183,7 +183,7 @@ private struct BitstreamReader {
         if case .char6 = element {
           // FIXME: Once the minimum deployment target bumps to macOS 11, use
           // the more ergonomic stdlib API everywhere.
-          if #available(macOS 11.0, *) {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please test

compnerd (approved):
Do we need to worry about tvOS and watchOS being covered?

abertelrud:
> Do we need to worry about tvOS and watchOS being covered?

Ideally yes.  I initially went down that route with this fix, just for completeness, but discovered that there are bigger blockers to building for those right now.  For example, while the posix_spawn APIs don't work iOS they are at least defined, but they are marked as unavailable for tvOS and watchOS.

So it would be some work to get TSC to work on tvOS and watchOS.  Since iOS is the one being actively used, that's the one that needs to continue working.  Perhaps we should have a clearer error message on watchOS and iOS until it's supported, if ever.  Ideally we'd have CI for at least building on iOS (only some of the unit tests run, since subprocess launching isn't available).

For the availability conditionals I've been taking the approach of copying the conditions from the definition to the call site, which is why I included tvOS and watchOS in that diff here.

PR #249, opened 2021-10-20
fix InterruptHandler to be safer

tomerd:
motivation: InterruptHandler can crash when signal is handled but globals are not initialized

changes:
* mark InterruptHandler as deprecated since DispatchSource provides this functionality already
* change InterruptHandler globals to static members
* add code to ensure the static state is initialized to make it more signal safe

rdar://76272674

### DIFF
--- a/Sources/TSCUtility/InterruptHandler.swift
+++ b/Sources/TSCUtility/InterruptHandler.swift
@@ -11,19 +11,19 @@
 import TSCLibc
 import TSCBasic
 
-/// Interrupt signal handling global variables
-private var wasInterrupted = false
-private var wasInterruptedLock = Lock()
-#if os(Windows)
-private var signalWatchingPipe: [HANDLE] = [INVALID_HANDLE_VALUE, INVALID_HANDLE_VALUE]
-#else
-private var signalWatchingPipe: [Int32] = [0, 0]
-private var oldAction = sigaction()
-#endif
-
 /// This class can be used by command line tools to install a handler which
 /// should be called when a interrupt signal is delivered to the process.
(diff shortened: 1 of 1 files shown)


Sources/TSCUtility/InterruptHandler.swift
```diff
@@ -11,19 +11,19 @@
...
-#if os(Windows)
-private var signalWatchingPipe: [HANDLE] = [INVALID_HANDLE_VALUE, INVALID_HANDLE_VALUE]
-#else
-private var signalWatchingPipe: [Int32] = [0, 0]
-private var oldAction = sigaction()
-#endif
-
 /// This class can be used by command line tools to install a handler which
 /// should be called when a interrupt signal is delivered to the process.
+@available(*, deprecated, message: "use DispatchSource instead")
```
CodaFi, line 16: 🎊

Sources/TSCUtility/InterruptHandler.swift
```diff
@@ -11,19 +11,19 @@
...
 public final class InterruptHandler {
+    /// Interrupt signal handling global variables
+    private static var wasInterrupted = false
+    private static var wasInterruptedLock = Lock()
+    #if os(Windows)
+    private static var signalWatchingPipe: [HANDLE] = [INVALID_HANDLE_VALUE, INVALID_HANDLE_VALUE]
+    #else
+    private static var signalWatchingPipe: [Int32] = [0, 0]
```
CodaFi, line 24: Probably better to use a tuple for this one and the windows one but meh that's really not important.

CodaFi (approved):
Thank you!

PR #251, opened 2021-10-29
Remove LinuxMain.swift

keith:
Swift 5.5+ Swift Package Manager warns that `--enable-test-discovery` is
enabled by default. I assume for this project that should mean this
isn't needed anymore on `main`.


tomerd (approved):

keith:
thanks!

PR #254, opened 2021-11-05
Correctly validate characters in semantic version identifiers

WowbaggersLiquidLunch:
Semantic Versioning 2.0.0 allows only ASCII alpha-numeric characters and "-" in identifiers.

This change makes version initialization more restrictive, so it might be source breaking.

Same as apple/swift-package-manager#3839, but for TSC's `Version`.

### DIFF
--- a/Sources/TSCUtility/Version.swift
+++ b/Sources/TSCUtility/Version.swift
@@ -28,7 +28,18 @@ public struct Version {
     /// The build metadata.
     public let buildMetadataIdentifiers: [String]
 
-    /// Creates a version object.
+    /// Creates a version with the provided components.
+    ///
+    /// - Parameters:
+    ///   - major: The major version number.
+    ///   - minor: The minor version number.
+    ///   - patch: The patch version number.
+    ///   - prereleaseIdentifiers: The pre-release identifier.
+    ///   - buildMetaDataIdentifiers: Build metadata that identifies a build.
+    ///
+    /// - Precondition: `major >= 0 && minor >= 0 && patch >= 0`.
+    /// - Precondition: `prereleaseIdentifiers` can conatin only ASCII alpha-numeric characters and "-".
+    /// - Precondition: `buildMetaDataIdentifiers` can conatin only ASCII alpha-numeric characters and "-".
     public init(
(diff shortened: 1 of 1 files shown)


WowbaggersLiquidLunch:
@swift-ci please smoke test

WowbaggersLiquidLunch:
@swift-ci please smoke test

tomerd:
@elsh please help review this one

Sources/TSCUtility/Version.swift
```diff
@@ -37,6 +48,18 @@ public struct Version {
...
+                $0.allSatisfy { $0.isASCII && ($0.isLetter || $0.isNumber || $0 == "-") }
+            },
+            #"Pre-release identifiers can contain only ASCII alpha-numeric characters and "-"."#
+        )
+        precondition(
+            buildMetadataIdentifiers.allSatisfy {
+                $0.allSatisfy { $0.isASCII && ($0.isLetter || $0.isNumber || $0 == "-") }
+            },
+            #"Build metadata identifiers can contain only ASCII alpha-numeric characters and "-"."#
+        )
```
tomerd, line 62: I am concerned about the use of preconditions here for what could be a runtime issue. should this be throwing instead?

WowbaggersLiquidLunch: This is a good point/suggestion. Making it throwing (or failable) also makes it possible to write tests against it.

Also, I found a few more bugs that I'm trying to fix before the 5.6 branch is cut: empty identifiers, leading zeros in numeric identifiers, and negative identifiers should not be allowed according to semver 2.0.0. With so many things to validate, it makes sense to throw errors for them instead of more lines of `precondition`s. 

One slight problem though is that the call sites of this initializer in SwiftPM (and possibly llbuild and swift-driver too) need to be updated when this initializer becomes throwing or failable, so there might need some deprecation dances.

WowbaggersLiquidLunch:
Changed this PR to a draft, because I'm going to do a new PR using throwing initializer as @tomerd suggested.

I'm leaving this draft open until the new PR is opened to gather additional reviews/suggestions.

PR #256, opened 2021-11-07
Remove iOS 13 deployment target

jakepetroules:
swift-tsc can build just fine with an iOS 9 deployment target, so let it
continue defaulting to that. Add the required availability annotations
in the unit tests to let the build succeed against the subset of APIs
that do require iOS 13.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -25,7 +25,6 @@ let package = Package(
     name: "swift-tools-support-core",
     platforms: [
         macOSPlatform,
-        .iOS(.v13)
     ],
     products: [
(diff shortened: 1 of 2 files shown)


jakepetroules:
This reverts part of #225, which was itself reverted in e932ca05d6c692a2ecb4c43e1149630444e28200, but not fully.

jakepetroules:
@swift-ci please test

tomerd:
cc @abertelrud

abertelrud (approved):
LGTM

PR #261, opened 2021-11-16
Add async version of `Process.popen()`

pcbeard:
This provides a new version of the `popen()` convenience method:

```swift
static public func popen(arguments: [String], environment: [String: String] = ProcessEnv.vars,
                             completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void)
```

This allows the caller to asynchronously receive a `ProcessResult` or `Swift.Error`. This was implemented to reduce the number of active threads necessary to execute multiple sub-processes simultaneously. A new unit test is provided that tests the API. In addition, enhancements were made to the `withTemporaryDirectory()` and `withTemporaryFile()` APIs to make them easier to use from asynchronous contexts:

```swift
public func withTemporaryFile<Result>(
  dir: AbsolutePath? = nil, prefix: String = "TemporaryFile", suffix: String = "", _ body: (TemporaryFile, @escaping (TemporaryFile) -> Void) throws -> Result
) throws -> Result

public func withTemporaryDirectory<Result>(
    dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory" , _ body: (AbsolutePath, @escaping (AbsolutePath) -> Void) throws -> Result
) throws -> Result
```

Both of these new functions pass the callback block a `@escaping` reference to a cleanup block, to allow the cleanup to be deferred until a later time. The old API uses a Bool argument to disable cleanup, and makes it the caller's responsibility to delete the temporary resources. The cleanup block reduces code duplication.

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -613,7 +613,7 @@ public final class Process {
 
             // Close the local write end of the output pipe.
             try close(fd: outputPipe[1])
-            
+
             // Create a thread and start reading the output on it.
             group.enter()
             let stdoutThread = Thread { [weak self] in
@@ -628,7 +628,7 @@ public final class Process {
                         }
                     }
                     group.leave()
-               } else if let stderrResult = (pendingLock.withLock { pending }) {
+                } else if let stderrResult = (pendingLock.withLock { pending }) {
                     // TODO: this is more of an error
                     self?.stateLock.withLock {
                         self?.state = .outputReady(stdout: .success([]), stderr: stderrResult)
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -230,7 +230,8 @@ public final class Process {
     // process execution mutable state
     private var state: State = .idle
     private let stateLock = Lock()
-    private static let stateQueue = DispatchQueue(label: "ProcessStateQueue")
+    private static let sharedCompletionQueue = DispatchQueue(label: "Process.sharedCompletionQueue")
+    private var completionQueue = Process.sharedCompletionQueue
 
     /// The result of the process execution. Available after process is terminated.
     /// This will block while the process is awaiting result
@@ -458,7 +459,7 @@ public final class Process {
             self.state = .readingOutput(sync: sync)
         }
 
-        group.notify(queue: Self.stateQueue) {
+        group.notify(queue: self.completionQueue) {
             self.stateLock.withLock {
                 self.state = .outputReady(stdout: .success(stdout), stderr: .success(stderr))
             }
@@ -719,7 +720,7 @@ public final class Process {
             completion(.failure(error))
         case .readingOutput(let sync):
             self.stateLock.unlock()
-            sync.notify(queue: Self.stateQueue) {
+            sync.notify(queue: self.completionQueue) {
                 self.waitUntilExit(completion)
             }
         case .outputReady(let stdoutResult, let stderrResult):
@@ -750,7 +751,7 @@ public final class Process {
                 stderrOutput: stderrResult
             )
             self.state = .complete(executionResult)
-            Self.stateQueue.async {
+            self.completionQueue.async {
                 self.waitUntilExit(completion)
             }
         }
@@ -827,9 +828,10 @@ extension Process {
     ///     will be inherited.
     /// - Returns: The process result.
     static public func popen(arguments: [String], environment: [String: String] = ProcessEnv.vars,
-                             completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
+                             queue: DispatchQueue? = nil, completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
         do {
             let process = Process(arguments: arguments, environment: environment, outputRedirection: .collect)
+            process.completionQueue = queue ?? Self.sharedCompletionQueue
             try process.launch()
             process.waitUntilExit(completion)
         } catch {
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -689,22 +689,22 @@ public final class Process {
     public func waitUntilExit() throws -> ProcessResult {
         let group = DispatchGroup()
         group.enter()
-        var processResult : ProcessResult?
-        var processError : Swift.Error?
-        self.waitUntilExit() { result, error in
+        var processResult : Result<ProcessResult, Swift.Error>?
+        self.waitUntilExit() { result in
             processResult = result
-            processError = error
             group.leave()
         }
         group.wait()
-        if let error = processError {
+        switch processResult! {
+        case .failure(let error):
             throw error
+        case .success(let result):
+            return result
         }
-        return processResult.unsafelyUnwrapped
     }
 
     /// Executes the process I/O state machine, calling completion block when finished.
-    private func waitUntilExit(_ completion: @escaping (ProcessResult?, Swift.Error?) -> Void) {
+    private func waitUntilExit(_ completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
         self.stateLock.lock()
         switch self.state {
         case .idle:
@@ -712,10 +712,10 @@ public final class Process {
             preconditionFailure("The process is not yet launched.")
         case .complete(let result):
             defer { self.stateLock.unlock() }
-            completion(result, nil)
+            completion(.success(result))
         case .failed(let error):
             defer { self.stateLock.unlock() }
-            completion(nil, error)
+            completion(.failure(error))
         case .readingOutputPipe(let sync):
             defer { self.stateLock.unlock() }
             sync.notify(queue: Self.stateQueue) {
@@ -826,7 +826,7 @@ extension Process {
     ///     will be inherited.
     /// - Returns: The process result.
     static public func popen(arguments: [String], environment: [String: String] = ProcessEnv.vars,
-                             completion: @escaping (ProcessResult?, Swift.Error?) -> Void) throws {
+                             completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) throws {
         let process = Process(arguments: arguments, environment: environment, outputRedirection: .collect)
         try process.launch()
         process.waitUntilExit(completion)
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -396,13 +396,14 @@ public final class Process {
         }
 
     #if os(Windows)
-        _process = Foundation.Process()
-        _process?.arguments = Array(arguments.dropFirst()) // Avoid including the executable URL twice.
-        _process?.executableURL = executablePath.asURL
-        _process?.environment = environment
+        let process = Foundation.Process()
+        _process = process
+        process.arguments = Array(arguments.dropFirst()) // Avoid including the executable URL twice.
+        process.executableURL = executablePath.asURL
+        process.environment = environment
 
         let stdinPipe = Pipe()
-        _process?.standardInput = stdinPipe
+        process.standardInput = stdinPipe
 
         let group = DispatchGroup()
 
@@ -446,8 +447,8 @@ public final class Process {
                 }
             }
 
-            _process?.standardOutput = stdoutPipe
-            _process?.standardError = stderrPipe
+            process.standardOutput = stdoutPipe
+            process.standardError = stderrPipe
         }
 
         // first set state then start reading threads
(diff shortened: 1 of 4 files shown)


pcbeard:
This PR depends on changes in https://github.com/apple/swift-corelibs-foundation/pull/3113 to function correctly on the Windows platform, which uses `Foundation.Process` under the hood.

Sources/TSCBasic/Process.swift
```diff
@@ -677,24 +688,40 @@ public final class Process {
...
+        case .failure(let error):
+            throw error
+        case .success(let result):
+            return result
+        }
+    }
+
+    /// Executes the process I/O state machine, calling completion block when finished.
+    private func waitUntilExit(_ completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
         self.stateLock.lock()
```
pcbeard, line 709: This could probably use a better name.

Sources/TSCBasic/Process.swift
```diff
@@ -180,10 +180,10 @@ public final class Process {
     // process execution mutable state
     private enum State {
         case idle
-        case readingOutputThread(stdout: Thread, stderr: Thread?)
         case readingOutputPipe(sync: DispatchGroup)
         case outputReady(stdout: Result<[UInt8], Swift.Error>, stderr: Result<[UInt8], Swift.Error>)
         case complete(ProcessResult)
+        case failed(Swift.Error)
```
tomerd, line 186: is there a different Error type in this code base that requires qualifying it as Swift.Error?

Sources/TSCBasic/Process.swift
```diff
@@ -677,24 +688,40 @@ public final class Process {
...
+        switch processResult! {
+        case .failure(let error):
+            throw error
+        case .success(let result):
+            return result
+        }
+    }
+
+    /// Executes the process I/O state machine, calling completion block when finished.
+    private func waitUntilExit(_ completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
```
tomerd, line 708: should we allow the caller to specify a queue instead of hard coding one? the caller may be some preferences

Sources/TSCBasic/Process.swift
```diff
@@ -677,24 +688,40 @@ public final class Process {
...
+            processResult = result
+            group.leave()
+        }
+        group.wait()
+        switch processResult! {
+        case .failure(let error):
+            throw error
+        case .success(let result):
+            return result
+        }
```
tomerd, line 704: can be simplified to `return try processResult.get()`

Sources/TSCBasic/Process.swift
```diff
@@ -230,6 +230,7 @@ public final class Process {
     // process execution mutable state
     private var state: State = .idle
     private let stateLock = Lock()
+    private static let stateQueue = DispatchQueue(label: "ProcessStateQueue")
```
tomerd, line 233: we should consider allowing the user to specify the queue (with a default). also this is a serial one, correct?

Sources/TSCBasic/Process.swift
```diff
@@ -677,24 +688,40 @@ public final class Process {
...
-            return result
-        case .readingOutputThread(let stdoutThread, let stderrThread):
-            self.stateLock.unlock() // unlock early since output read thread need to change state
-            // If we're reading output, make sure that is finished.
-            stdoutThread.join()
-            stderrThread?.join()
-            return try self.waitUntilExit()
+            completion(.success(result))
+        case .failed(let error):
+            defer { self.stateLock.unlock() }
```
tomerd, line 718: why defer and not unlock directly?

Sources/TSCBasic/Process.swift
```diff
@@ -677,24 +688,40 @@ public final class Process {
...
+            completion(.failure(error))
         case .readingOutputPipe(let sync):
-            self.stateLock.unlock() // unlock early since output read thread need to change state
-            sync.wait()
-            return try self.waitUntilExit()
+            defer { self.stateLock.unlock() }
```
tomerd, line 721: why defer and not unlock directly?

Sources/TSCBasic/Process.swift
```diff
@@ -648,22 +655,26 @@ public final class Process {
...
             } else {
                 pendingLock.withLock {
                     pending = .success([])  // no stderr in this case
                 }
             }
+            
             // first set state then start reading threads
             self.stateLock.withLock {
-                self.state = .readingOutputThread(stdout: stdoutThread, stderr: stderrThread)
+                self.state = .readingOutputPipe(sync: group)
```
tomerd, line 675: we should probably rename this to something else than readingOutputPipe if reusing it for the thread bases scenario as well, how about `readingOutput`

Sources/TSCBasic/TemporaryFile.swift
```diff
@@ -213,14 +237,14 @@ extension MakeDirectoryError: CustomNSError {
...
-///             If `body` has a return value, that value is also used as the
-///             return value for the `withTemporaryDirectory` function.
+///           If `body` has a return value, that value is also used as the
+///           return value for the `withTemporaryDirectory` function.
+///           The cleanup block should be called when the temporary directory is no longer needed.
 ///
 /// - Throws: MakeDirectoryError and rethrows all errors from `body`.
 public func withTemporaryDirectory<Result>(
-    dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory", removeTreeOnDeinit: Bool = false , _ body: (AbsolutePath) throws -> Result
+    dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory" , _ body: (AbsolutePath, @escaping (AbsolutePath) -> Void) throws -> Result
```
tomerd, line 247: can these changes move to a separate PR? or are they related to the async process?

tomerd (changes requested):
thanks @pcbeard this is something I wanted to do for a very long time <3

some feedback inline

Sources/TSCBasic/Process.swift
pcbeard: Yes, `ProcessResult.Error`.

Sources/TSCBasic/Process.swift
pcbeard: Possibly. Before we were using a .global queue to wake up the dispatch group. Using a shared serial queue seems better. But a shared specified queue might be fine too.

Sources/TSCBasic/Process.swift
pcbeard: Thanks for the tips. I'm still learning this pattern.

Sources/TSCBasic/Process.swift
pcbeard: Global/concurrent queues are in general discouraged as a dispatch anti-pattern. They can create extra threads unnecessarily. I needed a queue to field the .notify() calls on. It's only a callback when the state machine calls the completion block. We could even do away with `stateLock` possibly.

Sources/TSCBasic/Process.swift
pcbeard: Sounds good.

Sources/TSCBasic/TemporaryFile.swift
pcbeard: They are needed upstream in SPM.

Sources/TSCBasic/Process.swift
pcbeard: Fixed in e5f2e14.

Sources/TSCBasic/Process.swift
pcbeard: Fixed in e5f2e14.

Sources/TSCBasic/Process.swift
pcbeard: Added optional queue parameter in d67a7f0.

Sources/TSCBasic/Process.swift
pcbeard: Added optional queue parameter in d67a7f0.

Sources/TSCBasic/TemporaryFile.swift
tomerd: I am a bit iffy about the `@escaping (AbsolutePath) -> Void` cleanup argument passed into the closure. feel like the boolean cleanup is good enough for the public API and this should an internal implementation details?

pcbeard: My rationale for this is to avoid code duplication.

tomerd: @pcbeard right. not a big deal, mostly wondering if the variant with the cleanup closure needs to be public, or if the boolean variant is good enough as the public api. i think SwiftPM can just use the boolean variant?

tomerd (approved):

tomerd:
@swift-ci test

pcbeard:
Please hold off merging this until https://github.com/apple/swift-corelibs-foundation/pull/3113 is merged.

Sources/TSCBasic/Process.swift
```diff
@@ -230,6 +230,8 @@ public final class Process {
     // process execution mutable state
     private var state: State = .idle
     private let stateLock = Lock()
+    private static let sharedCompletionQueue = DispatchQueue(label: "Process.sharedCompletionQueue")
```
compnerd, line 233: Would be nice to use `org.swift.tools-support-core.process-completion` as the name for the queue.

Sources/TSCBasic/Process.swift
```diff
@@ -609,8 +613,9 @@ public final class Process {
 
             // Close the local write end of the output pipe.
             try close(fd: outputPipe[1])
-
+
```
compnerd, line 616: Bleeding whitespace

Sources/TSCBasic/Process.swift
```diff
@@ -622,11 +627,13 @@ public final class Process {
                             pending = readResult
                         }
                     }
-                } else if let stderrResult = (pendingLock.withLock { pending }) {
+                    group.leave()
+               } else if let stderrResult = (pendingLock.withLock { pending }) {
```
compnerd, line 631: Indentation seems off?

Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -42,20 +42,57 @@ class ProcessTests: XCTestCase {
     }
 
     func testPopen() throws {
+        #if os(Windows)
+        let echo = "echo.exe"
```
compnerd, line 46: This worries me - there is no `echo.exe` technically.  It is a MSys/2 binary that is part of git and if you add Git to your path, it might show up or it might now.  On the other hand, the vast majority of the t-s-c tests will just abort anyways, so whats a broken test?

The equivalent of echo is echo, which is a shell builtin.

Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -42,20 +42,57 @@ class ProcessTests: XCTestCase {
...
 
         // Test buffer larger than that allocated.
         try withTemporaryFile { file in
             let count = 10_000
             let stream = BufferedOutputByteStream()
             stream <<< Format.asRepeating(string: "a", count: count)
             try localFileSystem.writeFileContents(file.path, bytes: stream.bytes)
-            let outputCount = try Process.popen(args: "cat", file.path.pathString).utf8Output().count
+            #if os(Windows)
+            let cat = "cat.exe"
```
compnerd, line 60: Similar to echo, `cat.exe` is not a thing.  The equivalent to `cat` is `type` which is a shell builtin.

Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -42,20 +42,57 @@ class ProcessTests: XCTestCase {
...
+            #endif
+            let outputCount = try Process.popen(args: cat, file.path.pathString).utf8Output().count
             XCTAssert(outputCount == count)
         }
     }
 
+    func testPopenAsync() throws {
+        #if os(Windows)
+        let args = ["where.exe", "where"]
+        let answer = "C:\\Windows\\System32\\where.exe"
```
compnerd, line 72: There is no guarantee that the path is `C:\Windows\System32\where.exe` . The correct answer is `%SystemRoot%\System32\where.exe`.  You can use `GetSystemDirectoryW` to get the value for `%SystemRoot%` rather than expanding it via an environment variable.

Sources/TSCBasic/Process.swift
pcbeard: Fixed in f30c446.

Sources/TSCBasic/Process.swift
pcbeard: Fixed in 95e286b.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please test

PR #263, opened 2021-11-21
Async await popen

pcbeard:


Sources/TSCBasic/Process.swift
```diff
@@ -834,6 +834,20 @@ extension Process {
...
+    ///   - arguments: The arguments for the subprocess.
+    ///   - environment: The environment to pass to subprocess. By default the current process environment
+    /// - Returns: The process result.
+    /// - Throws: Any errors that cause the execution to fail.
+    @available(macOS 12.0, *)
+    static public func popen(arguments: [String], environment: [String : String] = ProcessEnv.vars) async throws -> ProcessResult {
+        try await withUnsafeThrowingContinuation {
+            Self.popen(arguments: arguments, environment: environment, completion: $0.resume)
+        }
+    }
```
tomerd, line 849: this would work, but still involves the an underlying dispatch queue. what would a "native" async/await implementation looks like? would it be faster and/or safer without the underlying dispatch queue?

Sources/TSCBasic/TemporaryFile.swift
```diff
@@ -263,6 +287,26 @@ public func withTemporaryDirectory<Result>(
     }
 }
 
+public func withTemporaryDirectory<Result>(
```
tomerd, line 290: useful! maybe material for a separate PR event without the async/await?

Sources/TSCBasic/Process.swift
pcbeard: I think we could refactor this so the use of the dispatch queue is at a different layer.

Tests/TSCBasicTests/ProcessTests.swift
```diff
@@ -92,6 +92,20 @@ class ProcessTests: XCTestCase {
             XCTFail()
         }
     }
+    
+    @available(macOS 12.0, *)
+    func testPopenAsyncAwait() async throws {
+        #if os(Windows)
+        let args = ["where.exe", "where"]
+        let answer = "C:\\Windows\\System32\\where.exe"
```
stevapple, line 100: You should not assume the path of `System` directory. Use `GetSystemDirectory` instead:
```suggestion
        var buffer = Array<WCHAR>(repeating: 0, count: Int(MAX_PATH + 1))
        guard GetSystemDirectoryW(&buffer, .init(buffer.count)) > 0 else {
            return XCTFail()
        }
        let answer = String(decodingCString: buffer, as: UTF16.self) + "\\where.exe"
```

stevapple (changes requested):

jakepetroules:
Closing as this functionality was added via #375. Please feel free to re-open or create a new PR if you have further contributions to propose!

PR #265, opened 2021-11-30
TSCBasic: remove `close` overload in extension

compnerd:
The `FileHandle` extension added an overloaded `close` that causes
ambiguity when using `close` on a `FileHandle`.  This manifested as a
build failure with apple/swift-package-manager#3888.  Remove the
function as swift-driver, swift-package-manager, and sourcekit-lsp are
able to build without this function (relying on Foundation to provide
it).

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -1016,9 +1016,5 @@ extension FileHandle: WritableByteStream {
     public func flush() {
         synchronizeFile()
     }
-
-    public func close() throws {
-        closeFile()
-    }
 }
 #endif


compnerd:
@swift-ci please test

compnerd:
CC: @abertelrud

abertelrud (approved):

abertelrud:
I just realized what happened here.  These were old covers that are no longer needed now that Foundation has refined its FileHandle API in Swift 5.0.

This was https://github.com/apple/swift-corelibs-foundation/commit/55c404ba3b7edac0f9712d5556bcd8f1d404e935

abertelrud:
That means that if this causes problems, I can also switch SwiftPM to use the deprecated `closeFile()` call.  But fixing forward is better.

abertelrud:
@compnerd Looks ready to go at this point?

PR #266, opened 2021-12-02
deprecate PkgConfig

tomerd:
motivation: PkgConfig was moved from SwiftPM into TSC, but really is closer to SwiftPM (not generic enough) and would be easier to maintain within SwiftPM itself

changes: mark PkgConfig and related utilities as deprecated


tomerd:
@swift-ci test

neonichu (approved):

PR #268, opened 2021-12-06
build: disable in-tree builds

compnerd:
This ensures that one does not accidentally build t-s-c in tree, which
means that the source tree is always pristine and the build tree root is
different from the source tree.  The previous change to ignore CMake
artifacts would only be caused by an accidental in-tree build which
should be frowned upon and be treated as unsupported.


compnerd:
@swift-ci please smoke test

neonichu (approved):

tomerd:
@swift-ci please test

tomerd (approved):

PR #269, opened 2021-12-09
TSCBasic: Make Process.waitUntilExit(_ completion:) public

simonjbeaumont:
The `Process` exposes a blocking `waitUntilExit()` which is implemented in terms of a non-blocking, private `waitUntilExit(_ completion:)`.

I'd like to be able to make a non-blocking wait from an asynchronous context in Swift >= 5.5.

I think providing an async-await API for `TSCBasic.Process` would be great, but in the meantime I'd like to be able to add the following extension in a dependent project:

```swift
#if swift(>=5.5)
@available(macOS 12.0, *)
extension TSCBasic.Process {
    @discardableResult
    public func awaitUntilExit() async throws -> ProcessResult {
        try await withCheckedThrowingContinuation { self.waitUntilExit($0.resume) }
    }
}
#endif
```

FWICT, `TSCBasic.waitUntilExit(_ completion:)` has locks in place so could be made public to support this effort. This PR does that.

### DIFF
--- a/Sources/TSCBasic/Process.swift
+++ b/Sources/TSCBasic/Process.swift
@@ -701,7 +701,7 @@ public final class Process {
     }
 
     /// Executes the process I/O state machine, calling completion block when finished.
-    private func waitUntilExit(_ completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
+    public func waitUntilExit(_ completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void) {
         self.stateLock.lock()
         switch self.state {
         case .idle:


tomerd:
cc @pcbeard wdyt?

tomerd:
@swift-ci test

compnerd (dismissed):
~Please explicitly test this on Windows; `async` has been somewhat unstable, so this should at least be limited to 5.6 IMO.~

Edit: nvm, this isn't adding the new `async` function, its exposing the internal version.

tomerd:
@compnerd yes - we are waiting with async for later in the maturity curve

simonjbeaumont:
@tomerd @compnerd yep, as noticed, this isn’t adding an `async` API, just allowing others to do so in downstream projects. 

WDYT?

compnerd:
@simonjbeaumont - given that the trees are closed while Apple is on holidays, I think that we can take another look after the repos are open again.

simonjbeaumont:
I assume that the repos are open for business again now? Just wanted to resurface this for someone's attention when they get a minute.

How would people feel about additionally adding the following API which would add the completion handler to the completion queue but not wait for exit, unlike `waitForExit(_:)`.

```swift
TSCBasic.Process.onExit(_ completion: @escaping (Result<ProcessResult, Swift.Error>) -> Void)
```

simonjbeaumont:
@compnerd @tomerd Friendly reminder for when you have time.

compnerd:
@simonjbeaumont hmm, any chance you could add some context why you want to make this public?  I think that overall, the process handling stuff in t-s-c has caused some amount of trouble, and this might encourage further use.

simonjbeaumont:
> @simonjbeaumont hmm, any chance you could add some context why you want to make this public?

I'm not sure how much more context I can give beyond the description—the code I want to add to _our_ project is in the PR description, and this relies on this PR. Essentially I need a non-blocking _wait for this process to exit_ function.

> I think that overall, the process handling stuff in t-s-c has caused some amount of trouble, and this might encourage further use.

Hm, I appreciate that. This project was suggested as having a process API that was a bit more suitable for our needs than `Foundation.Process`. Do you have any alternatives I could pursue?

simonjbeaumont:
@tomerd @compnerd did you have any more time to consider this?

tomerd:
@[simonjbeaumont I think the concern is not so much with making it public, but with the proposed usage pattern. using it in an async/await context would not be safe as it has some assumptions on threading model which may not be true in that context. what we would want here is to implement the process async methods (including this one) using idiomatic swift concurrency primitives (ie actors) and then expose those APIs publicly. if you want, a PR to do that is welcome

jakepetroules:
Closing as this functionality was added via #375. Please feel free to re-open or create a new PR if you have further contributions to propose!

PR #273, opened 2021-12-22
store lock files in temporary directory

tomerd:
motivation: local file system store lock files next to the file being locked and never cleans them up

changes:
* add a tempDirectory to FileSystem protocol
* update determineTempDirectory to use tempDirectory from localFileSystem
* move logic to handle local file system locks to FileLock and call it from localFileSystem
* expose an API on FileLock to create local file system locks, which by default uses the temp directory to store lock files, using the file-to-lock path and name to construct the lock file name
* move local file system lock logic to use the new FileLock API
* add tests

rdar://86644116

### DIFF
--- a/Sources/TSCBasic/Lock.swift
+++ b/Sources/TSCBasic/Lock.swift
@@ -44,9 +44,9 @@ extension ProcessLockError: CustomNSError {
         return [NSLocalizedDescriptionKey: "\(self)"]
     }
 }
-/// Provides functionality to aquire a lock on a file via POSIX's flock() method.
+/// Provides functionality to acquire a lock on a file via POSIX's flock() method.
 /// It can be used for things like serializing concurrent mutations on a shared resource
-/// by mutiple instances of a process. The `FileLock` is not thread-safe.
+/// by multiple instances of a process. The `FileLock` is not thread-safe.
 public final class FileLock {
 
     public enum LockType {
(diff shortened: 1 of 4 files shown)


tomerd:
https://github.com/apple/swift-package-manager/pull/3962

@swift-ci please test

jakepetroules (approved):

abertelrud (approved):

Sources/TSCBasic/Lock.swift
```diff
@@ -163,4 +167,24 @@ public final class FileLock {
...
+        // unless specified, we use the tempDirectory to store lock files
+        let lockFilesDirectory = lockFilesDirectory ?? localFileSystem.tempDirectory
+        if !localFileSystem.exists(lockFilesDirectory) {
+            throw FileSystemError(.noEntry, lockFilesDirectory)
+        }
+        if !localFileSystem.isDirectory(lockFilesDirectory) {
+            throw FileSystemError(.notDirectory, lockFilesDirectory)
+        }
+        // use the parent path to generate unique filename in temp
+        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory).appending(component: fileToLock.basename)).components.joined(separator: "_")
```
abertelrud, line 181: There's somewhat of a risk of overflow here since most operating systems / file systems have a much shorter file name limit than path limit.  I don't know if that will be an issue in practice — an alternate approach to consider here might be to hash all the but the last _n_ components of the path, which should still be indicative of what the lock file refers to.

tomerd: we can also re-root the path into temp a potential solution. would that be better than what we have here now?

PR #278, opened 2022-01-12
[5.6] FileLock encounters a runtime error when the path of the locked file is long (#277)

abertelrud:
This is cherry-picked from #277. 

**Explanation:**  FileLock constructs the name of the lock file based on the path of the file being locked.  When this path is very long, the name of the constructed lock file exceeds `NAME_MAX` (or more specifically the directory entry name length limit of the file system on which the directory that contains it resides, but `NAME_MAX` is a decent approximation on modern file systems commonly in use).

**Scope of Issue:**  This can affect any use of the file locking where the locked file is at a long path (i.e. deeply nested and/or with long path components).  This issue was seen in SwiftPM.

**Reason for Nominating to 5.6:**  This causes unit test failures in SwiftPM and can cause runtime failures in actual use (typically not crashes but errors such as build errors and other kinds of failures).  Although this can be worked around by moving the packages to a shallower path, it's not at all obvious that this is the problem nor is it always possibly to affect the paths of for example CI systems.

**Risk:**  Low — the problem is understood and the fix is straightforward.  The main issue would be if there is something special about Unicode content of paths that hasn't been accounted for.  Any failure to fix the problem should be evident from continued failure of unit tests such as in SwiftPM and its client IDEs.

**Reviewed By:**  @tomerd, @neonichu 

**Automated Testing:**  An updated unit test checks for this condition.

**Dependencies:**  None

**Impact on CI:**  None

**How to Verify:**  Create a package that uses this API and try to lock a path whose total length in UTF-8 bytes is greater than the value of `NAME_MAX` (this is also what the unit test does).

rdar://87471360

### DIFF
--- a/Sources/TSCBasic/Lock.swift
+++ b/Sources/TSCBasic/Lock.swift
@@ -178,7 +178,18 @@ public final class FileLock {
             throw FileSystemError(.notDirectory, lockFilesDirectory)
         }
         // use the parent path to generate unique filename in temp
-        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory).appending(component: fileToLock.basename)).components.joined(separator: "_") + ".lock"
+        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory)
+                                .appending(component: fileToLock.basename))
+                                .components.joined(separator: "_")
+                                .replacingOccurrences(of: ":", with: "_") + ".lock"
+#if os(Windows)
+        // NTFS has an ARC limit of 255 codepoints
+        var lockFileUTF16 = lockFileName.utf16.suffix(255)
+        while String(lockFileUTF16) == nil {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

abertelrud:
Linux: `No space left on device`

abertelrud:
@swift-ci please test linux

abertelrud:
Will need to resolve the Windows issue identified in https://github.com/apple/swift-tools-support-core/pull/280 before merging.  I really wish we had some kind of Windows PR testing to flag these issues.

abertelrud:
@swift-ci please test

tomerd (approved):

airspeedswift (approved):

PR #279, opened 2022-01-13
[Triple] More detailed errors if getting the triple from the compiler should fail

karwa:
This fails spuriously on Windows. It would be nice to get some more detailed output when that happens, and it's probably a useful thing for other platforms, too.

CC @compnerd


karwa:
See https://forums.swift.org/t/error-tscutility-triple-swift-fatal-error-could-not-determine-host-triple-malformed/47775

compnerd:
@swift-ci please test

karwa:
Yeah the comments are mostly just to break up what would otherwise be an unreadable glob of do-catches. They could also be blank lines, I suppose.

neonichu (approved):

PR #281, opened 2022-01-13
Fix a build failure on Windows, where `NAME_MAX` doesn't exist

abertelrud:


abertelrud:
@swift-ci test

compnerd (changes requested):
This still breaks, I think I have a fix, I'll upload it in a minute.

abertelrud:
> This still breaks, I think I have a fix, I'll upload it in a minute.

Thanks.  I think I just need to wrap the unit test in a condition too.

abertelrud:
@compnerd Conditionalized the unit test too.  That should be it.

abertelrud:
@swift-ci please test

abertelrud:
We do have other uses of `NAME_MAX`, but I didn't notice that they're inside big conditionals.

PR #282, opened 2022-01-13
TSCBasic: repair the Windows build after #277

compnerd:
This fixes multiple issues with the file locking implementation on
Windows.  `AbsolutePath.root` is meaningless on Windows - there is
no concept of a singular root on Windows, instead you have have 26
individual roots - A-Z.  Stripping the count of characters for the
root path leaves us with an invalid path string as the root drives
have been stripped.  The next character is an invalid character on
most file systems, and so we must replace `:`.  Instead of using a
`MAX_NAME` to compute a length of the ARC, enforce a limit for the
complete path (though an ARC would be limited to 255 codepoints on
NTFS).  This incidentally also corrects the emitted path, previous
path computation left us with the file in PWD.

### DIFF
--- a/Sources/TSCBasic/Lock.swift
+++ b/Sources/TSCBasic/Lock.swift
@@ -178,7 +178,18 @@ public final class FileLock {
             throw FileSystemError(.notDirectory, lockFilesDirectory)
         }
         // use the parent path to generate unique filename in temp
-        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory).appending(component: fileToLock.basename)).components.joined(separator: "_") + ".lock"
+        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory)
+                                .appending(component: fileToLock.basename))
+                                .components.joined(separator: "_")
+                                .replacingOccurrences(of: ":", with: "_") + ".lock"
+#if os(Windows)
+        // NTFS has an ARC limit of 255 codepoints
+        var lockFileUTF16 = lockFileName.utf16.suffix(255)
+        while String(lockFileUTF16) == nil {
+            lockFileUTF16 = lockFileUTF16.dropFirst()
+        }
+        lockFileName = String(lockFileUTF16) ?? lockFileName
+#else
         if lockFileName.hasPrefix(AbsolutePath.root.pathString) {
             lockFileName = String(lockFileName.dropFirst(AbsolutePath.root.pathString.count))
         }
@@ -191,6 +202,7 @@ public final class FileLock {
         }
         // we will never end up with nil since we have ASCII characters at the end
         lockFileName = String(lockFileUTF8) ?? lockFileName
+#endif
         let lockFilePath = lockFilesDirectory.appending(component: lockFileName)
 
         let lock = FileLock(at: lockFilePath)
--- a/Tests/TSCBasicTests/FileSystemTests.swift
+++ b/Tests/TSCBasicTests/FileSystemTests.swift
@@ -768,10 +768,10 @@ class FileSystemTests: XCTestCase {
 
             try _testFileSystemFileLock(fileSystem: localFileSystem, fileA: fileA, fileB: fileB, lockFile: lockFile)
 
-            // Test some long and edge case paths. We arrange to split between the C and the Cedilla if NAME_MAX is 255.
-            let longEdgeCase1 = tempDir.appending(component: String(repeating: "Façade!  ", count: Int(NAME_MAX)).decomposedStringWithCanonicalMapping)
+            // Test some long and edge case paths. We arrange to split between the C and the Cedilla by repeating 255 times.
+            let longEdgeCase1 = tempDir.appending(component: String(repeating: "Façade!  ", count: 255).decomposedStringWithCanonicalMapping)
             try localFileSystem.withLock(on: longEdgeCase1, type: .exclusive, {})
-            let longEdgeCase2 = tempDir.appending(component: String(repeating: "🏁", count: Int(NAME_MAX)).decomposedStringWithCanonicalMapping)
+            let longEdgeCase2 = tempDir.appending(component: String(repeating: "🏁", count: 255).decomposedStringWithCanonicalMapping)
             try localFileSystem.withLock(on: longEdgeCase2, type: .exclusive, {})
         }
     }


compnerd:
@swift-ci please test

compnerd:
@abertelrud I _think_ that this is mostly similar to #281 but missing the changes you found on the test suite, but it has the additional file name generation fix.  I think that we need part of your change still, a part of this change, and a part of another change to get this fully correct.

abertelrud:
> @abertelrud I _think_ that this is mostly similar to #281 but missing the changes you found on the test suite, but it has the additional file name generation fix. I think that we need part of your change still, a part of this change, and a part of another change to get this fully correct.

Thanks for this!  Right, at least the unit test's use of `NAME_MAX` needs to be conditioned out (I updated that PR while you were putting up this one).

compnerd:
I should be more explicit: I don't know how to do this M&A exercise - any thoughts on how to accomplish this the most effective way?

abertelrud:
I am not sure about the origins of the original `AbsolutePath.root` but it was there already.  So I guess there's a silver lining here in that there are some non-Windows-compatible runtime issues being addressed.

abertelrud:
> I should be more explicit: I don't know how to do this M&A exercise - any thoughts on how to accomplish this the most effective way?

I think we should proceed with this PR but just put in the conditional from #281 to skip that part of the unit test.

compnerd:
Once the dust from these changes settles, I believe that #278 also needs to be updated.

abertelrud:
> Once the dust from these changes settles, I believe that #278 also needs to be updated.

Yes, absolutely.

One thing about this PR is that I think you end up prepending the lockFilesDirectory twice.

Sources/TSCBasic/Lock.swift
```diff
@@ -178,13 +178,20 @@ public final class FileLock {
             throw FileSystemError(.notDirectory, lockFilesDirectory)
         }
         // use the parent path to generate unique filename in temp
-        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory).appending(component: fileToLock.basename)).components.joined(separator: "_") + ".lock"
+        var lockFileName = (resolveSymlinks(fileToLock.parentDirectory)
+                                .appending(component: fileToLock.basename))
+                                .components.joined(separator: "_")
+                                .replacingOccurrences(of: ":", with: "_") + ".lock"
+#if os(Windows)
+        var lockFileUTF8 = lockFilesDirectory.appending(component: lockFileName).pathString.utf8.suffix(Int(MAX_PATH))
```
abertelrud, line 186: This is supposed to be just the filename, so `lockFilesDirectory` gets prepended again below.

abertelrud: This means you'll end up with `lockFilesDirectory` twice.

Looking at this more, I think the part about `AbsolutePath.root` was try avoid pasting something that starts with `/` as a subpath.  I kept that part unmodified but it seems odd, since the file name should never have that in the first place.

My view is that we should just compute a SHA1 of the whole path, whatever it is, and then use that as the filename.  But it's true that it would be less readable.

abertelrud: Oh I see:  With non-Windows file paths, the `components` returns a `/` as the first path component.  So this is just stripping that out.

Taking a step back, I think what would make sense here would be to take just the components leading from the root of the path (whether that's `/` or `C:\` or whatever) and just using those as the basis for the filename.  All this is trying to do is to come up with a unique file name.

abertelrud: Is there an API to return just the path and not the root?

compnerd: That would be better, however, I _think_ that this new version should address the path issue.

compnerd: > Is there an API to return just the path and not the root?

I don't believe that there is such an API in the TSC Path API today.

Sources/TSCBasic/Lock.swift
```diff
@@ -178,7 +178,23 @@ public final class FileLock {
...
+        // NTFS has an ARC limit of 255 codepoints
+        var lockFileUTF16 = lockFileName.utf16.suffix(255)
+        while String(lockFileUTF16) == nil {
+            lockFileUTF16 = lockFileUTF16.dropFirst()
+        }
+        lockFileName = String(lockFileUTF16) ?? lockFileName
+
+        let path = lockFilesDirectory.appending(component: lockFileName)
+        if path.pathString.utf8.count > Int(MAX_PATH) {
+            lockFileName = String(lockFileName.dropFirst(path.pathString.utf8.count - Int(MAX_PATH)))
```
abertelrud, line 195: I don't think this will work since the count for `dropFirst()` is in characters but the count being checked is in UTF-8 bytes.

abertelrud: I kind of feel that the PATH_MAX check can probably be deferred — the other code path doesn't check for that either.

abertelrud: So just delete this second part for now, since that's generic anyway.

compnerd: Ah right - okay, let's make this entirely characters.  This should _mostly_ work right on Windows due to UTF16 (yes, surrogate pairs, blah, blah, blah, I said **mostly**).

compnerd: > So just delete this second part for now, since that's not actually platform specific anyway.

Sounds reasonable to me.  Better to be uniform.

abertelrud:
These new changes look good to me.  The only thing left is to conditionalize the unit test which references `NAME_MAX` (but doesn't really need to — it could just be 255).

abertelrud:
Then any kind of exceeding of `PATH_MAX` could be done in a separate PR.

compnerd:
@swift-ci please test

abertelrud:
Thanks for helping to deal with this and sorry it landed on you unexpected.  I didn't realize this would have a Windows effect, and I never was able to get Swift up and running in a Windows VM (it's not really feasible to manually test all changes in a separate Windows VM locally anyway)

compnerd:
I'd like to keep the authorship information for the changes to the test case (you noticed that and fixed it) which is why I haven't folded that into this change.

abertelrud:
> > I'd like to keep the authorship information for the changes to the test case (you noticed that and fixed it) which is why I haven't folded that into this change.
> 
> We need that, though, I think for this to pass? I certainly don't mind folding it in here, unless you specifically want to disavow it. :) But the fix is actually different — now that the Windows side is fixed too, through your changes, I think it makes sense to not conditionalize the unit test but just put the constant 255 instead of `NAME_MAX` there. Since it's just constructing test strings by repeating a phrase.

So what do you suggest as the next step here?  I think the only thing missing in this PR is `NAME_MAX` -> 255 in the unit test.

compnerd:
I don't mind at all if you push your bits of the changes to this change at all, I just care about the authorship information :)

abertelrud:
> I don't mind at all if you push your bits of the changes to this change at all, I just care about the authorship information :)

Ok, I see, so to check out your branch and add a commit to it?  Happy to do that.  I think the current CI will fail if we don't, right?  This PR still doesn't build on Windows IIUC?

compnerd:
I've built it on Windows locally, and been able to use the generated TSCBasic.dll to run swift-build, so I expect that this should be okay.

abertelrud:
> I've built it on Windows locally, and been able to use the generated TSCBasic.dll to run swift-build, so I expect that this should be okay.

Ok, sounds good.  Did you build the unit tests on Windows too?  Because I don't see how those would compile.

compnerd:
> Ok, sounds good. Did you build the unit tests on Windows too? Because I don't see how those would compile.

I did not, which is why I never noticed that.

abertelrud:
Ok, I added a commit to your branch in your fork, so the attribution information should be correct.

compnerd:
Thanks @abertelrud!

abertelrud:
Thanks @compnerd for your fixes here, and sorry about now having to restart the tests.  Meanwhile we should get @tomerd's signoff, and then I will update #278.

tomerd (approved):

PR #284, opened 2022-01-25
Fix build regression on iOS from #283

jakepetroules:
@compnerd Is this OK for Windows? Builds fine for macOS and iOS without warnings.

### DIFF
--- a/Sources/TSCBasic/misc.swift
+++ b/Sources/TSCBasic/misc.swift
@@ -162,8 +162,8 @@ public func exec(path: String, args: [String]) throws -> Never {
     guard execv(path, cArgs.cArray) != -1 else {
         throw SystemError.exec(errno, path: path, args: args)
     }
-    fatalError("unreachable")
   #endif
+    fatalError("unreachable")
 }
 
 @_disfavoredOverload


jakepetroules:
@swift-ci please test

BenchR267 (approved):

compnerd:
No, it doesn't work due to the use of `_exit` before the actual `fatalError` :-(

Sources/TSCBasic/misc.swift
```diff
@@ -162,8 +162,8 @@ public func exec(path: String, args: [String]) throws -> Never {
     guard execv(path, cArgs.cArray) != -1 else {
         throw SystemError.exec(errno, path: path, args: args)
     }
-    fatalError("unreachable")
   #endif
+    fatalError("unreachable")
```
compnerd, line 166: This shouldn't follow on Windows due to the use of `_exit`.

compnerd (changes requested):

Sources/TSCBasic/misc.swift
jakepetroules: Ah, I see @neonichu opened another PR (#285), let's go with that one.

PR #286, opened 2022-02-02
add isReadable and isWritable APIs to FileSystem

tomerd:
motivation: users often need to check if a path is readable and / or writable

changes:
* add isReadable and isWritable APIs to FileSystem
* implement in relevant FileSystem implementations
* add tests

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -162,6 +162,12 @@ public protocol FileSystem: AnyObject {
     /// Check whether the given path is accessible and is a symbolic link.
     func isSymlink(_ path: AbsolutePath) -> Bool
 
+    /// Check whether the given path is accessible and readable.
+    func isReadable(_ path: AbsolutePath) -> Bool
+
+    /// Check whether the given path is accessible and writable.
+    func isWritable(_ path: AbsolutePath) -> Bool
+
     // FIXME: Actual file system interfaces will allow more efficient access to
     // more data than just the name here.
     //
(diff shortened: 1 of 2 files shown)


tomerd:
https://github.com/apple/swift-package-manager/pull/4077

@swift-ci test

tomerd:
apple/swift-package-manager#4077

@swift-ci test

abertelrud (approved):

yim-lee (approved):

weissi:
@tomerd / @jakepetroules this broke the API and thus should've been tagged as 0.3.0 not 0.2.5

tomerd:
@weissi @jakepetroules was not aware there is any semantic versioning done on this library! SwiftPM, Driver and LSP which are the main consumers both depend on it in a branch-revision-locked way (🙄 I know) so there isn't much effort put into evolving APIs in a clean semantic way. fwiw my goal is to deprecate this library all together and not continue to invest in it

@jakepetroules you seem to have been the one tagging this library? would be good to better understand to what end its being tagged, and how to better coordinate if such tagging is in fact needed and the implications on versioning and Api evolution

weissi:
> @weissi @jakepetroules was not aware there is any semantic versioning done on this library! 

Well, it's 0.x so SemVer says "yolo", ie. you can say it's okay to break the API between 0.2.4 and 0.2.5. However, usually if you do do a breaking change, you'd increment the minor (ie. 0.2.4  -> 0.3).

> SwiftPM, Driver and LSP which are the main consumers both depend on it in a branch-revision-locked way (🙄 I know) so there isn't much effort put into evolving APIs in a clean semantic way. fwiw my goal is to deprecate this library all together and not continue to invest in it

Totally cool, I'm not arguing for us trying to not break the API, I'm arguing for using a new minor version for breaking changes (and not just a patch bump).


> @jakepetroules you seem to have been the one tagging this library? would be good to better understand to what end its being tagged, and how to better coordinate if such tagging is in fact needed and the implications on versioning and Api evolution

IMHO we should try to bump the minor for breaking changes but I don't think we need to have strict rules for API evolution. Breaking is fine by me but we should make that fact visible through version numbers.

tomerd:
this sounds fine, but I was not even aware its being tagged, so we need better coordination here

PR #294, opened 2022-03-04
TSCBasic: enable proper rendering on Windows

compnerd:
This restores the behaviour of rendering on Windows after a very long
time.  However, since `TERM` is not an environment variable on Windows,
simply elide the check for that.  This finally gives us proper rendering
on Windows.

### DIFF
--- a/Sources/TSCBasic/TerminalController.swift
+++ b/Sources/TSCBasic/TerminalController.swift
@@ -112,15 +112,13 @@ public final class TerminalController {
 
     /// Computes the terminal type of the stream.
     public static func terminalType(_ stream: LocalFileOutputByteStream) -> TerminalType {
-#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
This depends on https://github.com/apple/swift-package-manager/pull/4192 to get proper rendering.

compnerd:
@swift-ci please test

tomerd (approved):
is there a test we can add?

compnerd:
I think that is part of the problem, I don't have a good way to test this in a test case.  The behaviour is not something that is entirely in our control.  It requires synchronization between a test application, OS, llbuild, and s-p-m.  On this own this is just undoing the blanket "do not support rendering terminals".  This was a workaround for the underlying problem which has now been resolved.

neonichu (approved):

PR #298, opened 2022-03-19
TSCBasic: rework relative path construction on Windows

compnerd:
When constructing a path relative to a root, compose the final path
before canonicalization as the canonicalization may require lookups
in the base location.

We additionally pay a small penalty to convert the slashes to native
separators.  This is purely for aesthetic purposes, it ensures that the
path is computed initially rather than delayed until use.

This in conjunction with apple/swift-package-manager#4237 enables the
use of header search paths in SPM which are based in other targets.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -75,7 +75,29 @@ public struct AbsolutePath: Hashable {
         if PathImpl(string: str).isAbsolute {
             self.init(str)
         } else {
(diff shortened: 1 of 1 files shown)


millenomi:
cc @tomerd This is the second part of the fix in https://github.com/apple/swift-package-manager/pull/4237

tomerd:
@swift-ci please test

tomerd (approved):

PR #308, opened 2022-04-01
[5.6] TSCBasic: rework relative path construction on Windows (#298)

compnerd:
When constructing a path relative to a root, compose the final path
before canonicalization as the canonicalization may require lookups
in the base location.

We additionally pay a small penalty to convert the slashes to native
separators.  This is purely for aesthetic purposes, it ensures that the
path is computed initially rather than delayed until use.

This in conjunction with apple/swift-package-manager#4237 enables the
use of header search paths in SPM which are based in other targets.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -75,7 +75,29 @@ public struct AbsolutePath: Hashable {
         if PathImpl(string: str).isAbsolute {
             self.init(str)
         } else {
+#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @airspeedswift @ddunbar
Nomination for 5.6

abertelrud (approved):

airspeedswift (approved):

PR #310, opened 2022-04-11
TSCBasic: normalize the path before canonicalization

compnerd:
Normalize the path before canonicalization as it may fail to properly
canonicalize otherwise.  Furthermore, this ensures that we always
create the normalized path string.


compnerd:
@swift-ci please test

compnerd:
This should be associated with apple/swift-driver#1084

tomerd (approved):

PR #312, opened 2022-05-01
TSCBasic: correct path componentization on Windows

compnerd:
When converting a path to components on Windows, we need to split on
both `/` and `\`.  Instead, prefer to normalize the path and split using
the Foundation extension which will already perform the necessary
operations properly and is well tested.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -544,7 +544,10 @@ private struct UNIXPath: Path {
     //
     var components: [String] {
 #if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

compnerd:
This has an associated change in apple/swift-driver#1091 (will have to do a second test for SPM as well)

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

tomerd (approved):

PR #314, opened 2022-05-05
TSCBasic: change spelling for `root`

compnerd:
Unlike Unix, Windows does not have a singular unified file system root.
Instead, you have a forest with up to 26 roots (A-Z).  Due to the
current design limitations of the Path API, use a drive-relative path as
a root to terminate recursion when using the `InMemoryFileSystem` which
is required to work towards enabling tests for SPM.


compnerd:
@swift-ci please test

compnerd:
This doesn’t have a drive specified, and uses a relative path.  This is actually one of the most interesting cases of this.  The `InMemoryFileSystem` has no concept of the drive stack - that is, there’s nothing to emulate a drive with even.  Instead everything is relative - and because it attempts to encapsulate everything (intercepts all operations) and because it has a singular instance and root we can get away with it.

compnerd:
@swift-ci please test

neonichu (approved):

PR #315, opened 2022-05-07
TSCBasic: avoid validating the relative path on Windows

compnerd:
The generated relative path is un-anchored and thus cannot be validated
or canonicalized.  Introduce a constructor for the unsafe constructor of
a relative path and use that to create the relative path.  This enables
the construction of a non-canonical relative path for cousins.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -247,6 +247,13 @@ public struct RelativePath: Hashable {
         _impl = impl
     }
 
+    /// Private initializer for constructing a relative path without performing
+    /// normalization or canonicalization.  This will construct a path without
+    /// an anchor and thus may be invalid.
+    fileprivate init(unsafeUncheckedPath string: String) {
+        self.init(PathImpl(string: string))
+    }
+
     /// Initializes the RelativePath from `str`, which must be a relative path
     /// (which means that it must not begin with a path separator or a tilde).
     /// An empty input path is allowed, but will be normalized to a single `.`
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/TSCBasic/Path.swift
```diff
@@ -247,6 +247,10 @@ public struct RelativePath: Hashable {
         _impl = impl
     }
 
+    internal init(string: String) {
```
tomerd, line 250: does this need to be internal?

tomerd: also should we give it a better name, eg static method that points out it is unsafe etc?

compnerd: I think that we can get away with `fileprivate`.  I'll change that.  If you mean, should it be public?  I'm not sure if we want to encourage people to make unsafe paths.

I'm not sure I understand the bit on the "static method".  However, I can adjust this to something like `init(unsafeUncheckedPath:)`, though I'm not particularly happy with that the label.  If you have suggestions about the label, please share them.

compnerd:
@swift-ci please test

Sources/TSCBasic/Path.swift
compnerd: We may have to of course revisit that as we get further through the test suite in SPM (I think that I'm only ~80% through it).

tomerd: fileprivate and new spelling is good to address the concern

PR #316, opened 2022-05-08
TSCTestSupport: support more complex file systems

compnerd:
File systems are not guaranteed to be uniformly accessible through a
singular view.  Always use the current working directory to determine
the of the temporary directory as the path may not be viable for
conversion to a relative path.  This is mandatory for resolving
assumptions in the swift-package-manager test suite.


compnerd:
@swift-ci please test

compnerd:
I think that I might be able to get away without this.  Closing for now.

PR #319, opened 2022-05-12
TSCTestSupport: `SDKROOT` is required for Windows

compnerd:
The `SDKROOT` environment variable is used to compute the default value
for the `-sdk` parameter to the swift compiler which is required to
locate the standard library.  Do not filter the environment variable on
Windows as that will prevent the SPM test suite from succeeding.


compnerd:
@swift-ci please test

tomerd (approved):

PR #320, opened 2022-05-13
TSCBasic: adjust `AbsolutePath(_:relativeTo:)` on Windows

compnerd:
Normalise the path prior to concatenation and canonicalization.  This
enables the underlying API to properly perform this operation in a
single step and fixes a few cases where we would incorrect form paths.


compnerd:
@swift-ci please test

tomerd (approved):

PR #321, opened 2022-05-17
Create helper functions for temporary directories which supports new concurrency features

Overcot:
Motivation is to slowly adapt and new concurrency features which can be used in swiftpm


tomerd:
thanks for the PR @Overcot. we are looking to deprecate this library over time. as such, would you mind moving this to SwiftPM's Basics module?

Overcot:
> thanks for the PR @Overcot. we are looking to deprecate this library over time. as such, would you mind moving this to SwiftPM's Basics module?

Yeah, i did in PR https://github.com/apple/swift-package-manager/pull/5532

Overcot:
I guess i will close PR then

PR #323, opened 2022-05-21
TSCTestSupport: avoid dropping path components

compnerd:
Given that the `AbsolutePath` constructor invoked here is
`AbsolutePath(_:relativeTo:)` dropping the first component is not
meaningful for UNIX paths as an absolute path relative to root is going
be the same path as the path relative to the root being re-formed to be
relative to the root.  However, this meaningfully impacts paths on
Windows as the shared path representation currently does not include the
drive or server & share (in the UNC case) and the resulting absolute
path representation is a drive-relative path, where dropping the first
component alters the path representation.

### DIFF
--- a/Sources/TSCTestSupport/FileSystemExtensions.swift
+++ b/Sources/TSCTestSupport/FileSystemExtensions.swift
@@ -52,7 +52,7 @@ extension FileSystem {
         do {
             try createDirectory(root, recursive: true)
             for path in files {
-                let path = AbsolutePath(String(path.dropFirst()), relativeTo: root)
+                let path = AbsolutePath(path, relativeTo: root)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

tomerd:
done have tests to make sure this still works correctly on all platforms?

compnerd:
The tools-support-core tests don't work on Windows (I'm tempted to spend the necessary energy towards removing the dependency).  The SPM tests will definitely exercise this heavily, and the assertion in `AbsolutePath` (i.e. `path.pathString.utf8.first! == "/"`).  I _think_ that this should be sufficiently covered.  But, if there is something that you think might be useful to add - I can try to add something blind.

PR #324, opened 2022-05-21
TSCBasic: special case a few leading character cases for Path

compnerd:
The UNIX implementation special cases `/` and `~` which SPM depends on.
Handle the invalid file system character `~` specially on Windows as
well.  As we are now getting closer to normalized path spellings, we
need to add a special case for the leading path separator character.  On
UNIX, this signifies the root of the singular unified file system view.
However, Windows presents a forest to the user with 26 roots.
Consequently, a path headed by the path separator is a relative path
rather than an absolute path.  However, we treat the root-relative path
as an invalid relative path for the time being as the representation is
not flexible enough to account for this model.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -760,7 +760,9 @@ private struct UNIXPath: Path {
         defer { fsr.deallocate() }
 
         let realpath: String = String(cString: fsr)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

neonichu (approved):

PR #325, opened 2022-05-21
TSCBasic: protect against an invalid API usage

compnerd:
`filePathRepresentation` may not be invoked on an empty string.  This
protects against that case which does occur during the SPM test suite
execution.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -18,6 +18,11 @@ private typealias PathImpl = UNIXPath
 private typealias PathImpl = UNIXPath
 #endif
 
+#if canImport(Darwin)
+import Darwin.C
+#endif
+
+import class Foundation.NSURL
 import protocol Foundation.CustomNSError
 import var Foundation.NSLocalizedDescriptionKey
 
@@ -254,7 +259,18 @@ public struct RelativePath: Hashable {
     /// normalization or canonicalization.  This will construct a path without
     /// an anchor and thus may be invalid.
     fileprivate init(unsafeUncheckedPath string: String) {
-        self.init(PathImpl(string: string))
+        if string.isEmpty {
+            self.init(PathImpl(string: string))
+        } else {
+#if _runtime(_ObjC)
+            self.init(PathImpl(string: String(cString: NSURL(fileURLWithPath: string).fileSystemRepresentation)))
+#else
+            let normalized: UnsafePointer<Int8> = string.fileSystemRepresentation
+            defer { normalized.deallocate() }
+
+            self.init(PathImpl(string: String(cString: normalized)))
+#endif
+        }
     }
 
     /// Initializes the RelativePath from `str`, which must be a relative path


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test macOS platform

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test macOS platform

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

neonichu (approved):

tomerd:
this does a bit more than just the empty string case, can you elaborate on what is going on here?

cc @abertelrud

Sources/TSCBasic/Path.swift
```diff
@@ -254,7 +259,18 @@ public struct RelativePath: Hashable {
     /// normalization or canonicalization.  This will construct a path without
     /// an anchor and thus may be invalid.
     fileprivate init(unsafeUncheckedPath string: String) {
-        self.init(PathImpl(string: string))
+        if string.isEmpty {
```
tomerd, line 262: does the case need to be a guard and then return a ~const empty representation?

tomerd, line 262: does the case need to be a guard and then return a ~const empty representation?

compnerd:
@swift-ci please test macOS platform

tomerd:
@compnerd this does a bit more than just the empty string case, can you elaborate on what is going on here / why all the thing manipulation is needed?

cc @abertelrud @milseman

tomerd:
@compnerd this does a bit more than just the empty string case, can you elaborate on what is going on here / why all the thing manipulation is needed?

cc @abertelrud @milseman

compnerd:
@tomerd I don't follow, its doing nothing at all ...

`NSString.fileSystemRepresentation` cannot be invoked on an empty string, so the check (or guard if you really want it just spelt that way for your tastes) is just protecting against that.

If it is not empty, we normalize the string and return that because the path operations should be performed on the normalized form (otherwise we need to go in at each point where an API is called and re-normalize the path before each call, which is extremely expensive as strings must be re-encoded to UTF16, have the conversion, re-encoded to UTF8, manipulated and then re-encoded to UTF16 to perform the operation).

compnerd:
Hmm, wait, do you mean the `self.init(PathImpl(string: string))` in the case that the string is empty?  I don't see how that is any different than writing `self.init(PathImpl(string: ""))`.  We already know that `string` is empty, so its just passing in `string` as opposed to creating a new String value from the `""` literal.  Is that you prefer spelling that as `self.init(Self.constant)`?

tomerd:
iiuc the non-empty code path is now:

```
#if _runtime(_ObjC)
             self.init(PathImpl(string: String(cString: NSURL(fileURLWithPath: string).fileSystemRepresentation)))
 #else
             let normalized: UnsafePointer<Int8> = string.fileSystemRepresentation
             defer { normalized.deallocate() }

             self.init(PathImpl(string: String(cString: normalized)))
 #endif
```

that is quite different from that the code used to be, and the motivation for this change (why do we need to use fileSystemRepresentation? what problem is it solving) is not explained as the motivation for this PR

compnerd:
Ah, I see.

The difference in the ObjC runtime is because `fileSystemRepresentation` behaviour is different.  The reason for that is to give similar behaviour across the two paths.

As to why the call to `.fileSystemRepresentation` is to canonicalise the path (that will rewrite the filepath to use the natural path separator rather than POSIX.  This is important because subsequent use of the resulting path makes assumptions that the path is in the canonicalised form rather than POSIX.  If we do not perform this at this point, future attempts to get pieces of the path fail.

tomerd:
> As to why the call to .fileSystemRepresentation is to canonicalise the path (that will rewrite the filepath to use the natural path separator rather than POSIX. This is important because subsequent use of the resulting path makes assumptions that the path is in the canonicalised form rather than POSIX. If we do not perform this at this point, future attempts to get pieces of the path fail.

can you give some concrete examples of what this does exactly / what concrete problem it solves. I am a bit worried about replying on fileSystemRepresentation. 

@milseman  @abertelrud can you chime in here too?

milseman:
> @milseman @abertelrud can you chime in here too?

Sorry, I'm not familiar with what this is. This doesn't look to be using `FilePath`, should it be?

compnerd:
Right, this is still using the path utilities from tools-support-core, not the swift-system FilePath.

tomerd:
@compnerd iirc https://github.com/apple/swift-tools-support-core/pull/369 and https://github.com/apple/swift-package-manager/pull/5910 are designed to avoid some of these issues, we should prioritize getting them over the finish lines so we need less workarounds like this

compnerd:
@tomerd are you okay with merging this for now?  I'm going to likely take a pause on some of the SPM work and see if we can get the two mentioned changes merged as that is going to just be generally easier to make progress with.

tomerd:
@compnerd are you asking if we can merge https://github.com/apple/swift-tools-support-core/pull/369 and https://github.com/apple/swift-package-manager/pull/5910? I am happy to merge them if you can confirm they are not breaking anything for windows and helping with some of the windows path issues

compnerd:
@tomerd I meant merging this change so that I can have a little bit of breathing room to work through those two changes :)

compnerd:
@tomerd okay, can you please at least get the patch pair that you have building on Windows?  I should be able to put it through its paces on Windows once you have the patches complete.  Can we get that done early today so that I can test it?  Alternatively, can we merge this temporarily so that we can work through the issues in the much larger refactoring?

compnerd:
I believe that this can be closed off now that we have the other change merged.

PR #326, opened 2022-06-07
[Linux] Define _GNU_SOURCE when building TSCclibc

bnbarham:
Fixes missing declaration for `posix_spawn_file_actions_addchdir_np` on
Linux.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -125,4 +125,10 @@ let package = Package(
       .linkedLibrary("Pathcch", .when(platforms: [.windows])),
     ]
   }
+#elseif os(Linux)
+  if let TSCclibc = package.targets.first(where: { $0.name == "TSCclibc" }) {
+    TSCclibc.cSettings = [
+      .define("_GNU_SOURCE", .when(platforms: [.linux])),
+    ]
+  }
 #endif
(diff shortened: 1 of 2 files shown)


bnbarham:
@swift-ci please test

compnerd (approved):
This matches what I had experienced but failed to upstream; thank you for taking care of it.

tomerd:
thanks @bnbarham can you provide context to the fix? is there a test that is missing that could have prevented the omission?

bnbarham:
> thanks @bnbarham can you provide context to the fix? is there a test that is missing that could have prevented the omission?

It's currently failing on rebranch (corresponding to LLVM stable/20220421) - https://ci.swift.org/view/Swift%20rebranch/job/oss-swift-rebranch-pr-test-ubuntu-20_04/315. I didn't look too hard into what changed there, any ideas @compnerd?

tomerd:
@swift-ci smoke test

compnerd:
I don't remember what exposed this :(. This is the correct thing though - there's no guarantee that the interfaces are available - they are inherently non-portable and require opt-in.  I think that it might be that the importer doesn't define it now to correct other interactions?

tomerd (approved):

PR #327, opened 2022-06-08
make lock lock&unlock methods public for process migration to spm

Overcot:
As mentioned in https://github.com/apple/swift-package-manager/pull/5532#issuecomment-1133212593 there is a need to migrate some bits of `Process` to spm and in order to do that the `Lock.lock` and `Lock.unlock` are needed to be public

### DIFF
--- a/Sources/TSCBasic/Lock.swift
+++ b/Sources/TSCBasic/Lock.swift
@@ -19,11 +19,11 @@ public struct Lock {
     public init() {
     }
 
-    func lock() {
+    public func lock() {
         _lock.lock()
     }
 
-    func unlock() {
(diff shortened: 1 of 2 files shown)


Overcot:
> thanks @Overcot, since this is just a simple wrapper of `NSLock`, maybe we move it whole sale into SwiftPM Basics module?

Yes, i can do it. But would i need to remove the wrapper inside TSCBasic or leave it as it is?

tomerd:
> Yes, i can do it. But would i need to remove the wrapper inside TSCBasic or leave it as it is?

I suggest we leave it in, and put a `@available(*, deprecated)` on top of it

PR #334, opened 2022-06-25
Fix TSC diagnostic file parsing (handle parsing diagnostics with multiple blocks)

etcwilde:
This patch fixes parts of the .dia file reading. As originally
implemented, the reader would fail and emit an error if a diagnostic had
sub-diagnostics, i.e, a warning with notes.

Part of the issue was that the parser could only handle one block at a
time and couldn't handle the nesting of blocks. After fixing this, it
became apparent that there were some other issues. The parser was
consuming the records in a post-order traversal of the diagnostic graph,
when it really needs to be handling them in a pre-order ordering. This
wasn't an issue as a post-order and pre-ordering are the same if you can
only handle trees with a root node. Emitted in the wrong order, the
filemap, flagmap, and categorymap wouldn't be updated correctly,
resulting on dropping diagnostic information on the floor, when really
the information would come later.

In the end, I recorded the diagnostic records in the pre-order traversal
order into the `diagnosticRecords` property. When exiting from the last
active blockID in a diagnostic tree, I would run over the lists of
records, emitting a diagnostic for each. This way we emit diagnostics in
a pre-order ordering once we have collected all of the information for
the diagnostics in the tree.

Now, there's even more fun to be had. When constructing the diagnostic,
the code made an assumption that filename records, category records, and
flag records would come strictly before the diagnosticInfo, source
range, and fixIt records. This isn't the case. To work around that, I
run over the records, handling all the filenames, categories, and flags
first. Once those have been populated into their respective maps, we
iterate over the records again, handling the diagnosticInfo, source
ranges, and fixIts with the fully populated maps. This should prevent us
from dropping pieces of diagnostics on the ground, simply because
they're in a weird order.

Noticed this issue after posting https://github.com/apple/swift/pull/59136.


etcwilde:
@swift-ci please test

etcwilde:
@swift-ci please test

neonichu (approved):

abertelrud:
This is great — thank you @etcwilde!  One future improvement here would be to fold the nested annotations in under each `SerializedDiagnostics.Diagnostic` (as a new `SerializedDiagnostics.Diagnostic.Annotation` type) so that client code can reason about top-level diagnostics as discrete issues, with the "here is the location of the previous declaration" etc annotations attached to the top-level issues.  That can be done in a follow-on PR, though, and I can see if I can help sketch that out if you want.

PR #337, opened 2022-07-11
TSCUtility: support additional spellings for ARM

compnerd:
ARM uses `armv[4-9]-?[arm]` as the arch field to encode the ISA
version and the core type for ARMv7.  Use the custom parsing logic
similar to the other fields to accommodate the spelling.


compnerd:
@swift-ci please test

Sources/TSCUtility/Triple.swift
```diff
@@ -135,6 +141,39 @@ public struct Triple: Encodable, Equatable {
...
+            "armv6": .armv6,
+            "armv5": .armv5,
+            "arm": .arm,
+            "arm64": .arm64,
+            "arm64e": .arm64e,
+            "wasm32": .wasm32,
+        ]
+        if let match = candidates.first(where: { string.hasPrefix($0.key) })?.value {
+            if case let .armv7(core: _) = match {
+                if string.hasPrefix("armv7-a") || string.hasPrefix("armv7a") {
```
etcwilde, line 163: The init function won't accept `armv7-a` because it splits the triple into components delimited on `-`, (line 114) then you'll either read the `a` as the vendor, or emit an error because the triple has the wrong number of components.

etcwilde (commented):
I think this is a good start. I kind of wonder if it shouldn't be generalized more than just for the core? There are a bunch of sub-architectures under armv7 alone. It might make sense really to make the `arm` case take an optional ARM sub arch, which would then include `v7`, `v7a`, `v7s`, `v7k`, etc... Thoughts?

Sources/TSCUtility/Triple.swift
```diff
@@ -135,6 +143,43 @@ public struct Triple: Encodable, Equatable {
...
+            "amd64": .amd64,
+            "armv7": .armv7(core: nil),
+            "armv6": .armv6,
+            "armv5": .armv5,
+            "arm": .arm,
+            "arm64": .arm64,
+            "arm64e": .arm64e,
+            "wasm32": .wasm32,
+        ]
+        if let match = candidates.first(where: { string.hasPrefix($0.key) })?.value {
```
stevapple, line 163: Swift's `Dictionary` isn’t ordered, so we will run into random results here by using `first(where:)`.

PR #341, opened 2022-07-13
un-deprecate AbsolutePath::appending(RelativePath)

tomerd:
motivation: concatinating rlative path to absolute path is a key API we should preserve it, but improve the implementation

chnages: remove deprecation marker from AbsolutePath::appending(RelativePath)

work to improve the implementation is tracked in https://github.com/apple/swift-tools-support-core/issues/336

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -169,7 +169,6 @@ public struct AbsolutePath: Hashable {
     }
 
     /// Returns the absolute path with the relative path applied.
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci test

neonichu (approved):

compnerd:
I thought that the plan was to undeprecate on 5.7 only and change the interface on main?

tomerd:
@compnerd we prefer not to diverge across main and 5.7 that way, and would like to work on #336 immediately to get main to the desired state. the first step is #343, which we will follow up with a fix for #336 (slacked you separately about that)

PR #343, opened 2022-07-13
refactor windows path implementation out of UnixPath and into WindowsPath

tomerd:
motivation: every method on UnixPath has an #if os(Windows)

changes:
* define WindowsPath, peer to UnixPath
* move all windows code to WindowsPath, and all POSIX code to UnixPath
* set PathImpl type alias accordingly

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -470,7 +470,7 @@ extension Path {
 private struct WindowsPath: Path {
     let string: String
 
-    static let root = WindowsPath(string: "/")
+    static let root = Self(string: "/")
 
     static func isValidComponent(_ name: String) -> Bool {
         return name != "" && name != "." && name != ".." && !name.contains("/")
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci test

Sources/TSCBasic/Path.swift
```diff
@@ -467,24 +466,22 @@ extension Path {
     }
 }
 
-private struct UNIXPath: Path {
+#if os(Windows)
+private struct WindowsPath: Path {
     let string: String
 
-    static let root = UNIXPath(string: "/")
+    static let root = WindowsPath(string: "/")
```
compnerd, line 473: I think that this should be `\`.

```suggestion
    static let root = WindowsPath(string: "\\")
```
```suggestion
    static let root = WindowsPath(string: "/")
```

compnerd (approved):

compnerd:
We probably should ensure that this doesn't regress SPM ... at the very least, please do a cross-repo test for building the toolchain with this.

neonichu (approved):

Sources/TSCBasic/Path.swift
tomerd: I have not changed the logic here from the original code: https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCBasic/Path.swift#L473

tomerd:
@compnerd yes I kicked one off: https://github.com/apple/swift-package-manager/pull/5661

tomerd:
@swift-ci test

PR #347, opened 2022-08-19
TSCBasic: correct `resolvingSymlinks` on Windows

compnerd:
The function is not meant to resolve a symbolic link in the terminal arc
but rather anywhere along the path.  More specifically, it is meant to
behave similar to `realpath` on Unix.  The moral equivalent operation on
Windows is `GetFinalPathNameByHandle` which takes a handle to the file
and retrieves the actual location resolving any junctions in the path.
This repairs the behaviour of the function which is important for
SourceKit-LSP's test suite.


compnerd:
@swift-ci please test

neonichu (approved):

tomerd (approved):

PR #361, opened 2022-11-03
TSCBasic: mark `Path`-conforming types `Sendable`

MaxDesiatov:
These types only have a trivial `String` stored property in their implementation. Marking them as `Sendable` seems safe and allows them to be freely used in `async` contexts.


MaxDesiatov:
@swift-ci please test

neonichu (approved):

PR #362, opened 2022-11-03
Package.swift: bump `swift-tools-version` to 5.5

MaxDesiatov:
SwiftPM already requires 5.5, but now with `Sendable` introduced on some TSC types in #361, this is unlikely to compile without errors with Swift 5.4.


MaxDesiatov:
@swift-ci please test

neonichu (approved):

PR #363, opened 2022-11-04
Build process.c with _GNU_SOURCE

fweimer-rh:
Otherwise the build will fail on glibc 2.29 and later because posix_spawn_file_actions_addchdir_np is not declared.

A glibc change has been submitted as well, but it will not become available until glibc 2.37:

[PATCH] posix: Make posix_spawn extensions available by default https://sourceware.org/pipermail/libc-alpha/2022-November/143197.html

### DIFF
--- a/Sources/TSCclibc/process.c
+++ b/Sources/TSCclibc/process.c
@@ -1,5 +1,9 @@
 #if defined(__linux__)
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE /* for posix_spawn_file_actions_addchdir_np */
+#endif
+
 #include <errno.h>
 
 #include "process.h"


MaxDesiatov:
@swift-ci please test

neonichu (approved):

compnerd (changes requested):
This runs the risk of causing warnings such as `macro redefined`.  You should add a check around this or better yet, sink this into the build system.

compnerd:
IIRC it does if it is from the command line vs builtin vs in the source.  At least with CMake we can control it on a per source file level, which keeps the risk profile the same.  However, a `#ifndef` wrapper seems fine too.

compnerd:
I think that we do have squashes on this repo, so either is fine - it can be squashed before merging.

MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci please test

compnerd (approved):

PR #375, opened 2022-12-12
Add some noasync annotations and async overloads for improved Swift Concurrency interop

jakepetroules:


MaxDesiatov (approved):
Nice! 👍

tomerd:
how safe is this given all the custom locking and dispatch usage in TSC's Process?

tomerd:
can we also get some tests added

jakepetroules:
> how safe is this given all the custom locking and dispatch usage in TSC's Process?

Should be fine. We're just wrapping existing completion handler APIs in continuations. The unsafe part is using the existing APIs from Swift Concurrency, because they can block the calling thread, which is really bad.

> can we also get some tests added

Done.

jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

MaxDesiatov:
I always forget whether we have Windows CI set up here. If we so, looks like that job didn't run here.

PR #379, opened 2023-01-06
[Serialized diagnostics] Ignore unknown records.

DougGregor:
To improve forward compatibility, ignore any unknown records encountered during deserialization rather than throwing an error. libclang has behaved this way forever, and we should do the same.


DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

tomerd (approved):

abertelrud (approved):

compnerd (commented):
@DougGregor where is this used?  IIRC swift-driver has a copy of the Bitcode deserialization logic.

DougGregor:
> @DougGregor where is this used? IIRC swift-driver has a copy of the Bitcode deserialization logic.

SwiftPM has a little-used client of this (`swift package diagnose-api-breaking-changes`). swift-driver uses this version in some of its tests

PR #380, opened 2023-01-06
[Serialized diagnostics] Ignore unknown records.

DougGregor:
To improve forward compatibility, ignore any unknown records encountered during deserialization rather than throwing an error. libclang has behaved this way forever, and we should do the same.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

abertelrud (approved):

PR #382, opened 2023-01-18
TSCBasic: make FS-related value types Sendable

MaxDesiatov:
When using `FileSystem`-related code in `async` context we need these types to be `Sendable`.

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -111,9 +111,9 @@ public extension FileSystemError {
 }
 
 /// Defines the file modes.
-public enum FileMode {
+public enum FileMode: Sendable {
 
-    public enum Option: Int {
+    public enum Option: Int, Sendable {
         case recursive
         case onlyFiles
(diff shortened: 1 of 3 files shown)


Sources/TSCBasic/FileSystem.swift
```diff
@@ -122,17 +122,17 @@ public enum FileMode {
     case userWritable
     case executable
 
-    internal var setMode: (Int16) -> Int16 {
+    public func setMode(_ originalMode: Int16) -> Int16 {
```
MaxDesiatov, line 125: I wanted to pull in some of the FS-related code into SwiftPM, but that depends on this function being `public`. That was also a good opportunity to convert it from this unconventional `var`-returning-a-closure notation.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

tomerd (approved):

PR #392, opened 2023-02-07
[Wrangle] Revert noasync annotation.

nate-chandler:
The annotation is only available in Swift 5.8 and above.  Some nodes are still building with Swift 5.7.

Reverts the following:
- Revert "Require a 5.8 compiler to conditionally compile the noasync attribute"
- Revert "Add some noasync annotations and async overloads for improved Swift Concurrency interop"


nate-chandler:
@swift-ci please test

neonichu (approved):

PR #402, opened 2023-03-18
Make `FileSystem` and conforming types as `Sendable`

MaxDesiatov:
Types that conform to `FileSystem` either don't have in-memory state to be protected with locks at all and use blocking I/O under the hood (like `LocalFileSystem`), or already protect their state with locks, like `InMemoryFileSystem`. Few other types in SwiftPM conforming to `FileSystem` either also use locks (like `GitFileSystemView`) or are read-only (like `VirtualFileSystem`). I'm convinced that adding a `Sendable` requirement on `FileSystem` is beneficial overall, since it's already passed around between Dispatch queues and other concurrent code that would allow us to resolve sendability warnings.

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -146,7 +146,7 @@ public enum FileMode: Sendable {
 /// substitute a virtual file system or redirect file system operations.
 ///
 /// - Note: All of these APIs are synchronous and can block.
-public protocol FileSystem: AnyObject {
+public protocol FileSystem: AnyObject, Sendable {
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

weissi (approved):
looks legit to me!

neonichu (approved):

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test macos

PR #410, opened 2023-04-19
TSCBasic: make `LocalFileSystem` a struct

MaxDesiatov:
`LocalFileSystem` doesn't contain any storage for state, it should be safely convertible to `struct`. For this we have to remove `AnyObject` requirement from the `FileSystem` protocol. As a nice benefit, `@unchecked Sendable` conformance on `LocalFileSystem` can be removed as it now inherits `Sendable` conformance from `FileSystem`.

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -146,7 +146,7 @@ public enum FileMode: Sendable {
 /// substitute a virtual file system or redirect file system operations.
 ///
 /// - Note: All of these APIs are synchronous and can block.
-public protocol FileSystem: AnyObject, Sendable {
+public protocol FileSystem: Sendable {
     /// Check whether the given path exists and is accessible.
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

tomerd (approved):

MaxDesiatov:
Build failure due to SourceKit-LSP relying on this being a reference type and using `===` for an identity comparison:

```
SKCoreTests/ToolchainRegistryTests.swift:523:48: error: argument type 'any FileSystem' expected to be an instance of a class or class-constrained type
  precondition(!clang && !swiftc && !clangd || fs === localFileSystem || !shouldChmod,
```

@tomerd should we just remove that precondition there or is it better to keep it as a class and preserve that `AnyObject` constraint on the `FileSystem` protocol, effectively closing this PR?

tomerd:
I think testing like this for local file system is anti-pattern, this should be replaced with better logic

PR #412, opened 2023-04-21
[5.9] Add `hasAttribute` to `FileSystem`

MaxDesiatov:
Cherry-pick of https://github.com/apple/swift-tools-support-core/pull/414.

## Motivation

Detecting whether a given absolute path is quarantined is required for giving an actionable error message to SwiftPM users in case a Swift SDK bundle is quarantined.

## Modifications

This new function returns `true` if a given path has an attribute applied when the file system supports this attribute, and returns `false` if such attribute is not applied or it isn't supported.

## Risk

For TSC itself it's an additive change with low to no risk, as updated protocol requirement has a default implementation. This new requirement is also "standalone" as it isn't used in any other TSC function, so it doesn't have any impact on existing code.

rdar://107392863

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -138,10 +138,30 @@ public enum FileMode: Sendable {
 }
 
 /// Extended file system attributes that can applied to a given file path. See also ``FileSystem/hasAttribute(_:_:)``.
-public enum FileSystemAttribute: String {
+public enum FileSystemAttribute: RawRepresentable {
     #if canImport(Darwin)
-    case quarantine = "com.apple.quarantine"
+    case quarantine
     #endif
+
+    public init?(rawValue: String) {
+        switch rawValue {
+        #if canImport(Darwin)
+        case "com.apple.quarantine":
+            self = .quarantine
+        #endif
+        default:
+            return nil
+        }
+    }
+
+    public var rawValue: String {
+        switch self {
+        #if canImport(Darwin)
+        case .quarantine:
+            return "com.apple.quarantine"
+        #endif
+        }
+    }
 }
 
 // FIXME: Design an asynchronous story?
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

Sources/TSCBasic/FileSystem.swift
```diff
@@ -168,6 +168,10 @@ public protocol FileSystem: AnyObject {
     /// Check whether the given path is accessible and writable.
     func isWritable(_ path: AbsolutePath) -> Bool
 
+    /// Returns `true` if a given path has a quarantine attribute applied if when file system supports this attribute.
+    /// Returns `false` if such attribute is not applied or it isn't supported.
+    func hasQuarantineAttribute(_ path: AbsolutePath) -> Bool
```
compnerd, line 173: Does it make sense to spell this as `hasAttribute(_:)`?  That would allow checking for extended attributes Linux.

MaxDesiatov:
@swift-ci test macos

Sources/TSCBasic/FileSystem.swift
MaxDesiatov: I previously discussed this with @tomerd and @abertelrud and we ended up with this specialized version, but I wonder if making it slightly more generic in terms of supporting other attributes also makes sense?

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

compnerd (approved):

neonichu (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
Currently blocked by https://github.com/apple/swift-tools-support-core/pull/419

tomerd:
@swift-ci test

MaxDesiatov:
@swift-ci test linux

PR #413, opened 2023-04-21
TSCBasic: deprecate `<<<` operator

MaxDesiatov:
Since TSCBasic has become a kitchen sink for multiple types and functions, it's important to be able to selectively import those. It is done with `import struct`, `import func`, and similar statements. Unfortunately, this doesn't work for operators, which means you have to import the whole module when you need access just to a single operator. This inadvertently pollutes the environment of available symbols, which is undesirable.

By deprecating the operator and adding a `send(_:)` function on `WritableByteStream` we allow selectively importing this functionality without importing the whole module.

This also brings the API closer to Swift conventions, which in my understanding don't favor use of operators for stream I/O. Looks like `<<<` was initially brought over to TSC as a C++ idiom.

### DIFF
--- a/Sources/TSCBasic/DiagnosticsEngine.swift
+++ b/Sources/TSCBasic/DiagnosticsEngine.swift
@@ -148,11 +148,11 @@ public final class DiagnosticsEngine: CustomStringConvertible {
 
     public var description: String {
         let stream = BufferedOutputByteStream()
-        stream <<< "["
+        stream.send("[")
         for diag in diagnostics {
-            stream <<< diag.description <<< ", "
+            stream.send(diag.description).send(", ")
         }
-        stream <<< "]"
+        stream.send("]")
         return stream.bytes.description
     }
 }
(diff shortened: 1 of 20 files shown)


MaxDesiatov:
@cit

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci clean test

MaxDesiatov:
@swift-ci test macos

tomerd (commented):
is it time to deprecate ByteStream? this predates much of swift maturing on non-DArwin platforms and I am not sure it is still required

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
> is it time to deprecate ByteStream? this predates much of swift maturing on non-DArwin platforms and I am not sure it is still required

I hope it is, but it's used in a bunch of places in SwiftPM like logging that relies on `ThreadSafeOutputByteStream`. Would we want a replacement of that type to rely on Swift Concurrency instead and become an actor? Would logging become `async` then?

neonichu (approved):

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test macos

PR #423, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

### DIFF
--- a/Sources/TSCUtility/InterruptHandler.swift
+++ b/Sources/TSCUtility/InterruptHandler.swift
@@ -8,6 +8,8 @@
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+#if !canImport(Musl)
+
 import TSCLibc
 import TSCBasic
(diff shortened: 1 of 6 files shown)


MaxDesiatov:
@swift-ci test

neonichu (approved):

Sources/TSCUtility/InterruptHandler.swift
```diff
@@ -8,6 +8,8 @@
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+#if !canImport(Musl)
+
 import TSCLibc
 import TSCBasic
```
MaxDesiatov, line 15: Declarations in this file are deprecated and we're not going to use them when building for Musl.

PR #426, opened 2023-07-10
TSCTestSupport: account for `:` in function names

compnerd:
Windows does not permit `:` in the file names.  Substitute `:` with `_` to permit tests to execute on Windows.


compnerd:
@swift-ci please test

MaxDesiatov (approved):

PR #431, opened 2023-09-18
Fix platform conditions in `Package.swift`

MaxDesiatov:
Current use of `#if os` is not compatible with cross-compilation.


MaxDesiatov:
@swift-ci test

neonichu (approved):
Thanks, seems ok since we do not support 5.3 anymore.

PR #432, opened 2023-09-18
[5.10] Fix platform conditions in `Package.swift`

MaxDesiatov:
Cherry-pick of #431.

Current use of `#if os` is not compatible with cross-compilation.


MaxDesiatov:
@swift-ci test

neonichu (approved):

compnerd (approved):
Thanks for cleaning that up!  It never was great but we couldn't get them in due to compatibility issues.

PR #444, opened 2023-11-24
Android: add better nullability checks for nullability annotations added in NDK 26

finagolfin:
This is needed because [Bionic recently added a bunch of these annotations](https://android.googlesource.com/platform/bionic/+/00a3dbaab65a7d628312fd67099c662942b1d45f%5E%21/#F0). I made sure this pull doesn't break anything by testing it on linux x86_64 and with the previous NDK 25c also. I used this patch with others to build the Swift toolchain for my Android CI, finagolfin/swift-android-sdk#122, and [the Termux app for Android](https://github.com/termux/termux-packages/commit/d10fd452662d3c39c0f19a901433bce8711b4688#diff-b8f3b540c354923485cb06eba2d66237536f89e88fd51e170495ddd8f5d32622R208), which now uses NDK 26b.

### DIFF
--- a/Sources/TSCBasic/FileSystem.swift
+++ b/Sources/TSCBasic/FileSystem.swift
@@ -490,8 +490,7 @@ private struct LocalFileSystem: FileSystem {
 
     func readFileContents(_ path: AbsolutePath) throws -> ByteString {
         // Open the file.
-        let fp = fopen(path.pathString, "rb")
-        if fp == nil {
+        guard let fp = fopen(path.pathString, "rb") else {
             throw FileSystemError(errno: errno, path)
         }
         defer { fclose(fp) }
@@ -520,8 +519,7 @@ private struct LocalFileSystem: FileSystem {
 
     func writeFileContents(_ path: AbsolutePath, bytes: ByteString) throws {
         // Open the file.
-        let fp = fopen(path.pathString, "wb")
-        if fp == nil {
+        guard let fp = fopen(path.pathString, "wb") else {
             throw FileSystemError(errno: errno, path)
         }
         defer { fclose(fp) }
--- a/Sources/TSCBasic/WritableByteStream.swift
+++ b/Sources/TSCBasic/WritableByteStream.swift
@@ -790,7 +790,7 @@ public final class LocalFileOutputByteStream: FileOutputByteStream {
     override final func writeImpl(_ bytes: ArraySlice<UInt8>) {
         bytes.withUnsafeBytes { bytesPtr in
             while true {
-                let n = fwrite(bytesPtr.baseAddress, 1, bytesPtr.count, filePointer)
+                let n = fwrite(bytesPtr.baseAddress!, 1, bytesPtr.count, filePointer)
                 if n < 0 {
                     if errno == EINTR { continue }
                     errorDetected(code: errno)
--- a/Sources/TSCTestSupport/PseudoTerminal.swift
+++ b/Sources/TSCTestSupport/PseudoTerminal.swift
@@ -24,7 +24,7 @@ public final class PseudoTerminal {
         if openpty(&primary, &secondary, nil, nil, nil) != 0 {
             return nil
         }
-        guard let outStream = try? LocalFileOutputByteStream(filePointer: fdopen(secondary, "w"), closeOnDeinit: false) else {
+        guard let outStream = try? LocalFileOutputByteStream(filePointer: fdopen(secondary, "w")!, closeOnDeinit: false) else {
             return nil
         }
         self.outStream = outStream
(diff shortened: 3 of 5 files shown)


Sources/TSCBasic/WritableByteStream.swift
```diff
@@ -790,7 +790,11 @@ public final class LocalFileOutputByteStream: FileOutputByteStream {
     override final func writeImpl(_ bytes: ArraySlice<UInt8>) {
         bytes.withUnsafeBytes { bytesPtr in
             while true {
+              #if os(Android)
+                let n = fwrite(bytesPtr.baseAddress!, 1, bytesPtr.count, filePointer)
```
finagolfin, line 794: This force-unwrap works fine on linux too, but I separated it out anyway in case there is concern it breaks other platforms.

Sources/TSCTestSupport/PseudoTerminal.swift
```diff
@@ -24,7 +24,7 @@ public final class PseudoTerminal {
         if openpty(&primary, &secondary, nil, nil, nil) != 0 {
             return nil
         }
-        guard let outStream = try? LocalFileOutputByteStream(filePointer: fdopen(secondary, "w"), closeOnDeinit: false) else {
+        guard let outStream = try? LocalFileOutputByteStream(filePointer: fdopen(secondary, "w")!, closeOnDeinit: false) else {
```
finagolfin, line 27: I didn't bother separating this out into a `guard let` since it's just for the tests, can do so if wanted.

Sources/TSCBasic/FileSystem.swift
```diff
@@ -491,8 +491,8 @@ private struct LocalFileSystem: FileSystem {
 
     func readFileContents(_ path: AbsolutePath) throws -> ByteString {
         // Open the file.
-        let fp = fopen(path.pathString, "rb")
-        if fp == nil {
+        let fpo = fopen(path.pathString, "rb")
+        guard let fp = fpo else {
```
compnerd, line 495: Why not merge these two lines?

```suggestion
        guard let fp = fopen(path.pathString, "rb") else {
```

Sources/TSCBasic/FileSystem.swift
```diff
@@ -521,8 +521,8 @@ private struct LocalFileSystem: FileSystem {
 
     func writeFileContents(_ path: AbsolutePath, bytes: ByteString) throws {
         // Open the file.
-        let fp = fopen(path.pathString, "wb")
-        if fp == nil {
+        let fpo = fopen(path.pathString, "wb")
+        guard let fp = fpo else {
```
compnerd, line 525:
```suggestion
        guard let fp = fopen(path.pathString, "wb") else {
```

Sources/TSCBasic/Process.swift
```diff
@@ -147,6 +147,9 @@ public final class Process {
 
         /// The current OS does not support the workingDirectory API.
         case workingDirectoryNotSupported
+
+        /// The stdin could not be opened.
+        case stdinNotOpening
```
compnerd, line 152: What about `stdinUnavailable`?  "not opening" is not particularly descriptive, and makes you question what happened.

Sources/TSCBasic/FileSystem.swift
finagolfin: Good point, will change.

Sources/TSCBasic/WritableByteStream.swift
finagolfin: I see this done elsewhere in this repo so should be fine, doing the force unwrap on all platforms now.

Sources/TSCBasic/FileSystem.swift
finagolfin: Done.

Sources/TSCBasic/FileSystem.swift
finagolfin: Done.

Sources/TSCBasic/Process.swift
finagolfin: Done.

finagolfin:
Rebased and made the changes asked for, plus fixed an Android test.

neonichu (approved):

finagolfin:
@kateinoigakukun, can I get a CI run here?

kateinoigakukun:
@swift-ci Please test

Sources/TSCBasic/WritableByteStream.swift
glessard: `ArraySlice` currently never passes a buffer with a nil `baseAddress`. It's arguably an implementation detail, but we rely on it in many places.

finagolfin: OK, I only modified this because the compiler complained and errored. Is there something you'd like me to change here?

glessard: No, I just wanted to further clarify why the force unwrap is correct here.

PR #445, opened 2023-11-25
build: allow building in Debug mode on Windows

compnerd:
With this tweak, we should be able to build the package in debug mode
even on Windows.  We always use the release mode DLL form of the C/C++
runtime on Windows and can build the Swift code in Debug or Release mode
optimizations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,21 @@
+
+if(POLICY CMP0091)
+  cmake_policy(SET CMP0091 NEW)
+endif()
+
 cmake_minimum_required(VERSION 3.19)
 
 project(SwiftTSC LANGUAGES C Swift)
 
 set(CMAKE_Swift_LANGUAGE_VERSION 5)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+# TODO: lift this restriction once we have a split Swift runtime build which can
+# be built with a debug C runtime.
+set(CMAKE_Swift_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
 
+# TODO: lift this restriction once we have a split Swift runtime build which can
+# be built with a debug C runtime.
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

CMakeLists.txt
```diff
@@ -1,10 +1,17 @@
...
+
 cmake_minimum_required(VERSION 3.19)
 
 project(SwiftTSC LANGUAGES C Swift)
 
 set(CMAKE_Swift_LANGUAGE_VERSION 5)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+set(CMAKE_Swift_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
 
+set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
```
etcwilde, line 14: Why is this `MultiThreadedDLL` and not `MultiThreaded$<$<CONFIG:Debug>:Debug>DLL`?
Don't we want it to expand to the library that matches, or are we not using the debug DLL in either case?

Same for the above.

etcwilde: ah, okay, so just tying the C portions to release builds so it's just a {static,dynamic}{debug,release}Swift x release{static,dynamic}C/C++ runtime bits. okay.

etcwilde: Do we need to mix the release and debug Swift with release and debug C/C++? Wouldn't it make more sense to have everything be debug or everything be release? And would it make sense to limit the debug builds to either just dll's/static-linked? Then we only need to build the debug half for both either statically or dynamically and it's no longer a combinatorial explosion. Then the debug build is only one configuration and each thing gets built three times.

compnerd: I think that if we want to reduce it further we should continue to not support static linking.  We could do dynamic only with debug and release mode and match Swift and C/C++.

Issue #446, opened 2023-11-25
case sensitivity of `Path` environment variable on windows


cobbal:
Trying to run swift from emacs on windows (niche use case, I know) I was running into an error about swift not being able to find "swift-test.exe". Checking environment variables, I think I've tracked this down to a combination of emacs using unconventional capitalization, and a case sensitivity inside of SPM.

When run from cmd.exe, the environment variable is `Path=C:\...`, but from emacs it's `PATH=C:\...`

From https://github.com/apple/swift-tools-support-core/blob/4d539ff4c3d75394a78e17f68b675ad82434dd71/Sources/TSCBasic/ProcessEnv.swift#L55-L62
it looks like swift is looking for the "Path" one and not doing the typical-on-windows case insensitive lookup.

This brings me finally to a question: Is this a swift bug and/or an emacs bug? As far as I understand, while "Path" is the official casing, environment variables are expected to be treated insensitively on windows (which works a bit unfortunately with swift's representation as `[String: String]`)

I can create a PR to fix this, and I may file a bug against emacs as well, but I wanted to check first, since my windows experience is pretty limited.

compnerd:
I would say it is a bug on both sides.  emacs should use `Path`, however, Swift should be cognizant of the case insensitivity as well.

compnerd:
@cobbal not sure if this actually fixes the emacs issue, but is definitely a step in the right direction.  Unfortunately, we do need to audit the rest of the stack to handle this :(.

PR #447, opened 2023-11-25
TSCBasic: support case insensitivity for environment

compnerd:
Windows is case insensitive for the environment control block, which we did not honour.  This causes issues when Swift is used with programs which are incorrectly cased (e.g. emacs).  Introduce an explicit wrapper type for Windows to make the lookup case insensitive, canonicalising the name to lowercase.  This allows us to treat `Path` and `PATH` identically (along with any other environment variable and case matching) which respects the Windows behaviour.  Additionally, migrate away from the POSIX variants which do not handle the case properly to the Windows version which does.

Fixes: #446

### DIFF
--- a/Sources/TSCBasic/Process/Process.swift
+++ b/Sources/TSCBasic/Process/Process.swift
@@ -53,7 +53,12 @@ public struct ProcessResult: CustomStringConvertible, Sendable {
     public let arguments: [String]
 
     /// The environment with which the process was launched.
-    public let environment: [String: String]
+    public let environmentBlock: ProcessEnvironmentBlock
+
+    @available(*, deprecated, renamed: "env")
+    public var environment: [String:String] {
+        Dictionary<String, String>(uniqueKeysWithValues: self.environmentBlock.map { ($0.key.value, $0.value) })
+    }
 
     /// The exit status of the process.
     public let exitStatus: ExitStatus
@@ -71,7 +76,7 @@ public struct ProcessResult: CustomStringConvertible, Sendable {
     /// See `waitpid(2)` for information on the exit status code.
     public init(
         arguments: [String],
-        environment: [String: String],
+        environmentBlock: ProcessEnvironmentBlock,
         exitStatusCode: Int32,
         normal: Bool,
         output: Result<[UInt8], Swift.Error>,
@@ -92,23 +97,58 @@ public struct ProcessResult: CustomStringConvertible, Sendable {
             exitStatus = .terminated(code: WEXITSTATUS(exitStatusCode))
         }
       #endif
-        self.init(arguments: arguments, environment: environment, exitStatus: exitStatus, output: output,
-            stderrOutput: stderrOutput)
+        self.init(arguments: arguments, environmentBlock: environmentBlock, exitStatus: exitStatus, output: output, stderrOutput: stderrOutput)
+    }
+
+    @available(*, deprecated, message: "use `init(arguments:environmentBlock:exitStatusCode:output:stderrOutput:)`")
+    public init(
+        arguments: [String],
+        environment: [String:String],
+        exitStatusCode: Int32,
+        normal: Bool,
+        output: Result<[UInt8], Swift.Error>,
+        stderrOutput: Result<[UInt8], Swift.Error>
+    ) {
+        self.init(
+            arguments: arguments,
+            environmentBlock: .init(environment),
+            exitStatusCode: exitStatusCode,
+            normal: normal,
+            output: output,
+            stderrOutput: stderrOutput
+        )
     }
 
     /// Create an instance using an exit status and output result.
     public init(
         arguments: [String],
-        environment: [String: String],
+        environmentBlock: ProcessEnvironmentBlock,
         exitStatus: ExitStatus,
         output: Result<[UInt8], Swift.Error>,
         stderrOutput: Result<[UInt8], Swift.Error>
     ) {
         self.arguments = arguments
-        self.environment = environment
+        self.environmentBlock = environmentBlock
         self.output = output
         self.stderrOutput = stderrOutput
         self.exitStatus = exitStatus
+    }
+
+    @available(*, deprecated, message: "use `init(arguments:environmentBlock:exitStatus:output:stderrOutput:)`")
+    public init(
+        arguments: [String],
+        environment: [String:String],
+        exitStatus: ExitStatus,
+        output: Result<[UInt8], Swift.Error>,
+        stderrOutput: Result<[UInt8], Swift.Error>
+    ) {
+        self.init(
+            arguments: arguments,
+            environmentBlock: .init(environment),
+            exitStatus: exitStatus,
+            output: output,
+            stderrOutput: stderrOutput
+        )
     }
 
     /// Converts stdout output bytes to string, assuming they're UTF8.
@@ -245,14 +285,19 @@ public final class Process {
     /// The current environment.
     @available(*, deprecated, message: "use ProcessEnv.vars instead")
     static public var env: [String: String] {
-        return ProcessInfo.processInfo.environment
+        ProcessEnv.vars
     }
 
     /// The arguments to execute.
     public let arguments: [String]
 
     /// The environment with which the process was executed.
-    public let environment: [String: String]
+    @available(*, deprecated, message: "use `environmentBlock` instead")
+    public var environment: [String:String] {
+        Dictionary<String, String>(uniqueKeysWithValues: environmentBlock.map { ($0.key.value, $0.value) })
+    }
+
+    public let environmentBlock: ProcessEnvironmentBlock
 
     /// The path to the directory under which to run the process.
     public let workingDirectory: AbsolutePath?
@@ -323,21 +368,41 @@ public final class Process {
     ///     continue running even if the parent is killed or interrupted. Default value is true.
     ///   - loggingHandler: Handler for logging messages
     ///
-    @available(macOS 10.15, *)
     public init(
         arguments: [String],
-        environment: [String: String] = ProcessEnv.vars,
+        environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block,
         workingDirectory: AbsolutePath,
         outputRedirection: OutputRedirection = .collect,
         startNewProcessGroup: Bool = true,
         loggingHandler: LoggingHandler? = .none
     ) {
         self.arguments = arguments
-        self.environment = environment
+        self.environmentBlock = environmentBlock
         self.workingDirectory = workingDirectory
         self.outputRedirection = outputRedirection
         self.startNewProcessGroup = startNewProcessGroup
         self.loggingHandler = loggingHandler ?? Process.loggingHandler
+    }
+
+    @_disfavoredOverload
+    @available(macOS 10.15, *)
+    @available(*, deprecated, renamed: "init(arguments:environmentBlock:workingDirectory:outputRedirection:startNewProcessGroup:loggingHandler:)")
+    public convenience init(
+        arguments: [String],
+        environment: [String:String] = ProcessEnv.vars,
+        workingDirectory: AbsolutePath,
+        outputRedirection: OutputRedirection = .collect,
+        startNewProcessGroup: Bool = true,
+        loggingHandler: LoggingHandler? = .none
+    ) {
+        self.init(
+            arguments: arguments,
+            environmentBlock: .init(environment),
+            workingDirectory: workingDirectory,
+            outputRedirection: outputRedirection,
+            startNewProcessGroup: startNewProcessGroup,
+            loggingHandler: loggingHandler
+        )
     }
 
     /// Create a new process instance.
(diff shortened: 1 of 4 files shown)


compnerd:
@swift-ci please test

Sources/TSCBasic/Process.swift
```diff
@@ -23,6 +23,13 @@ import Dispatch
 
 import _Concurrency
 
+#if os(Windows)
+public typealias ProcessEnvironmentBlock = Dictionary<CaseInsensitiveString, String>
+#else
+public typealias ProcessEnvironmentBlock = Dictionary<String, String>
+#endif
```
MaxDesiatov, line 30: This will require #if os(Windows)` at the place of use. How about creating a new `ProcessEnvironment` type hiding these platform differences under the hood?

compnerd: I'm not sure I follow.  The typealias is meant to be the means for avoiding the `#if os` check at the use site.

compnerd: Ah, I see what you mean now.  Sure, that sounds like a good idea.

compnerd: Do you want me to keypath the type so that we can easily expose the other methods on the dictionary?

MaxDesiatov: Do you mean `@dynamicMemberLookup` with key paths or something else? IIUC `@dynamicMemberLookup` doesn't support proxying methods, only effectless properties. Maybe this new type could provide a read-only access to the underlying dictionary instead, while modifications are allowed through a custom subscript doing the normalization itself?

compnerd: Ah, bleh; okay.  In that case, what we have currently is reasonable.  We can extend this later if needed.

compnerd:
@swift-ci please test

Sources/TSCBasic/Process.swift
```diff
@@ -23,6 +23,30 @@ import Dispatch
 
 import _Concurrency
 
+public struct ProcessEnvironmentBlock {
```
MaxDesiatov, line 26: Not sure if `Block` brings much if anything valuable to the naming. Also, this should include a doc comment.
```suggestion
/// Windows is case insensitive for the environment control block. This caused issues when Swift is used with 
/// programs which are incorrectly cased (e.g. emacs).  This type Introduces an explicit wrapper type for 
/// Windows to make the lookup case insensitive, canonicalising the name to lowercase.  This allows us to treat 
/// `Path` and `PATH` identically (along with any other environment variable and case matching) which respects /// the Windows behaviour.
public struct ProcessEnvironment {
```

Sources/TSCBasic/Process.swift
```diff
@@ -23,6 +23,30 @@ import Dispatch
...
+#endif
+
+    public init<S: Sequence>(uniqueKeysWithValues keysAndValues: S)
+            where S.Element == (Key, String) {
+        storage.init(uniqueKeysWithValues: keysAndValues)
+    }
+
+    public subscript(_ key: String) -> String {
+#if os(Windows)
+        return stroage[CaseInsensitiveString(key)]
```
MaxDesiatov, line 42:
```suggestion
        return storage[CaseInsensitiveString(key)]
```

Sources/TSCBasic/Process.swift
MaxDesiatov: Now looking at `ProcessEnv`, why not actually move this over there and call it `ProcessEnv.Storage`?

MaxDesiatov: That's fair, but IMO the doc comment suggestion still holds.

compnerd:
@swift-ci please test

Sources/TSCBasic/Process.swift
```diff
@@ -23,6 +23,34 @@ import Dispatch
 
 import _Concurrency
 
+/// Windows is case insensitive for the environment control block. This caused issues when Swift is used with 
+/// programs which are incorrectly cased (e.g. emacs).  This type Introduces an explicit wrapper type for 
+/// Windows to make the lookup case insensitive, canonicalising the name to lowercase.  This allows us to treat 
+/// `Path` and `PATH` identically (along with any other environment variable and case matching) which respects /// the Windows behaviour.
+public struct ProcessEnvironmentBlock {
+#if os(Windows)
+    public typealias Key = CaseInsensitiveString
```
MaxDesiatov, line 32: One last question: if we have `public subscript(_ key: String) -> String` written as is, do we need to have this `typealias` at all? Why not keep `CaseInsensitiveString` internal instead of `public`?

compnerd: The constraint on the initialiser would force us to make it public.  Now ... maybe if I can make the initialiser internal ... 🤔

compnerd: Bleh, we still need the type to be public as it is used in ProcessEnv.swift as the type of a public variable.

Sources/TSCBasic/ProcessEnv.swift
```diff
@@ -9,25 +9,74 @@
...
+#if os(Windows)
+    public static var vars: [CaseInsensitiveString: String] { _vars }
+    private static var _vars: Dictionary<CaseInsensitiveString, String> = {
+        .init(uniqueKeysWithValues: ProcessInfo.processInfo.environment.map {
+            (CaseInsensitiveString($0), $1)
+        })
+    }()
+#else
     public static var vars: [String: String] { _vars }
     private static var _vars = ProcessInfo.processInfo.environment
```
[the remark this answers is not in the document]
compnerd: Hmm, I don't know about the deprecation, but at least the replacement is what I was experimenting with.

MaxDesiatov:
@swift-ci test

compnerd:
This will need a complementary change in swift-driver, I'll whip that up in a bit.

MaxDesiatov:
It may break SwiftPM and/or SourceKit-LSP too, that's why deprecation of the old API and introducing a new one may be preferable.

compnerd:
Sure, the deprecation is something to consider, but this is a problem for Windows and needs to propagate.

compnerd:
@swift-ci please test

compnerd:

compnerd:
@MaxDesiatov I'm no longer convinced that trying to do this is valuable and we should just expose this in all its nuance to all users.  We will end up having to rebuild the full `Dictionary` interface otherwise.  Either way, the constraints that are needed on the public interfaces are such that we need to make `CaseInsensitiveString` public in any case, so there is no true value in trying to convert this to a custom data structure.

Sources/TSCBasic/Process.swift
```diff
@@ -53,7 +144,7 @@ public struct ProcessResult: CustomStringConvertible, Sendable {
     public let arguments: [String]
 
     /// The environment with which the process was launched.
-    public let environment: [String: String]
+    public let environment: ProcessEnvironmentBlock
```
MaxDesiatov, line 147: This is a breaking API change, while new API surface has to be introduced in additive manner, with older API deprecated first. Otherwise this will break clients of TSC that we don't have direct control over.

MaxDesiatov (changes requested):

MaxDesiatov:
One more request here: this needs a test case to clarify and verify all of the new API surface, since the original bug report doesn't provide a self-contained reproducible test case. Hence my confusion about desired behaviour in a different version of the change.

compnerd:
I don't know about tests that would be useful here - this is providing the API surface that is necessary to build the dependencies - I don't think that this makes sense to provide tests for unless we just give up on the type representation and just force people to spell out the dictionary type.

MaxDesiatov:
We have to verify that the new API surface works well in the first place. Tests codify a specification for the behavior and the API shape.

For example, if you want to preserve string case sensitivity internally while allowing case insensitive lookup at the same time, as you mentioned in the other PR, tests should verify that this property holds, and will also guarantee that it doesn't regress in the future.

Sources/TSCBasic/ProcessEnv.swift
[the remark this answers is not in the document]
compnerd: There, introduced the deprecated interfaces.

Sources/TSCBasic/Process.swift
compnerd: Done

Sources/TSCBasic/Process/ProcessEnv.swift
```diff
@@ -11,23 +11,77 @@
...
+      #endif
+    }
+
     /// Returns a dictionary containing the current environment.
-    public static var vars: [String: String] { _vars }
-    private static var _vars = ProcessInfo.processInfo.environment
+    public static var block: ProcessEnvironmentBlock { _vars }
+
+    private static var _vars =
+        ProcessEnvironmentBlock(uniqueKeysWithValues: ProcessInfo.processInfo.environment.map { (ProcessEnvironmentBlock.Key($0.key), $0.value) })
```
MaxDesiatov, line 69: formatting nit for line length
```suggestion
    private static var _vars = ProcessEnvironmentBlock(
        uniqueKeysWithValues: ProcessInfo.processInfo.environment.map { 
            (ProcessEnvironmentBlock.Key($0.key), $0.value) 
        }
    )
```

Sources/TSCBasic/Process/ProcessEnv.swift
```diff
@@ -11,23 +11,77 @@
...
+
+#if os(Windows)
+public typealias ProcessEnvironmentBlock = [CaseInsensitiveString:String]
+extension ProcessEnvironmentBlock {
+  public init(_ dictionary: [String:String]) {
+    self.init(uniqueKeysWithValues: dictionary.map { (CaseInsensitiveString($0.key), $0.value) })
+  }
+}
+#else
+public typealias ProcessEnvironmentBlock = [String:String]
```
MaxDesiatov, line 48: I'm unsure if this is a good name, since AFAIU this is specific to Windows NT. Is there something more generic we could use instead?

Sources/TSCBasic/Process/ProcessEnv.swift
```diff
@@ -11,23 +11,77 @@
...
-    private static var _vars = ProcessInfo.processInfo.environment
+    public static var block: ProcessEnvironmentBlock { _vars }
+
+    private static var _vars =
+        ProcessEnvironmentBlock(uniqueKeysWithValues: ProcessInfo.processInfo.environment.map { (ProcessEnvironmentBlock.Key($0.key), $0.value) })
 
     /// Invalidate the cached env.
     public static func invalidateEnv() {
-        _vars = ProcessInfo.processInfo.environment
+        _vars = ProcessEnvironmentBlock(uniqueKeysWithValues: ProcessInfo.processInfo.environment.map { (CaseInsensitiveString($0.key), $0.value) })
```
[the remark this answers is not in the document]
compnerd: I don't think it does, I don't see any deprecation warnings.  Note that I left the backing store named `_vars`.  I can change it to `_block` if you prefer.

Sources/TSCBasic/Process/ProcessEnv.swift
compnerd: We already have a mixture of `env` and `environment`, I think that this is a reasonable alternative.

Sources/TSCBasic/Process/Process.swift
```diff
@@ -323,21 +376,34 @@ public final class Process {
     ///     continue running even if the parent is killed or interrupted. Default value is true.
     ///   - loggingHandler: Handler for logging messages
     ///
+    public init(arguments: [String], environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block, workingDirectory: AbsolutePath, outputRedirection: OutputRedirection = .collect, startNewProcessGroup: Bool = true, loggingHandler: LoggingHandler? = .none) {
```
MaxDesiatov, line 379: formatting nit for long lines
```suggestion
    public init(
        arguments: [String], 
        environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block, 
        workingDirectory: AbsolutePath, 
        outputRedirection: OutputRedirection = .collect, 
        startNewProcessGroup: Bool = true, 
        loggingHandler: LoggingHandler? = .none
    ) {
```

Sources/TSCBasic/Process/Process.swift
```diff
@@ -900,36 +994,48 @@ extension Process {
     ///   - environment: The environment to pass to subprocess. By default the current process environment
     ///     will be inherited.
     ///   - loggingHandler: Handler for logging messages
-    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
-    static public func popen(
-        arguments: [String],
-        environment: [String: String] = ProcessEnv.vars,
-        loggingHandler: LoggingHandler? = .none
-    ) async throws -> ProcessResult {
+    static public func popen(arguments: [String], environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block, loggingHandler: LoggingHandler? = .none) async throws -> ProcessResult {
```
MaxDesiatov, line 997:
```suggestion
    static public func popen(
        arguments: [String], 
        environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block, 
        loggingHandler: LoggingHandler? = .none
    ) async throws -> ProcessResult {
```

Sources/TSCBasic/Process/Process.swift
```diff
@@ -900,36 +994,48 @@ extension Process {
...
+    }
+
     /// Execute a subprocess and returns the result when it finishes execution
     ///
     /// - Parameters:
     ///   - args: The arguments for the subprocess.
     ///   - environment: The environment to pass to subprocess. By default the current process environment
     ///     will be inherited.
     ///   - loggingHandler: Handler for logging messages
+    static public func popen(args: String..., environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block, loggingHandler: LoggingHandler? = .none) async throws -> ProcessResult {
```
MaxDesiatov, line 1026:
```suggestion
    static public func popen(
        args: String...,
        environmentBlock: ProcessEnvironmentBlock = ProcessEnv.block,
        loggingHandler: LoggingHandler? = .none
    ) async throws -> ProcessResult {
```

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

MaxDesiatov (approved):
This LGTM now, thank you for iterating on this 👍

compnerd:
@swift-ci please test

compnerd:

PR #451, opened 2023-12-12
Add `ProcessEnvironment` type to replace `ProcessEnv`

MaxDesiatov:
`ProcessEnvironment` through subscript access guarantees that its `storage` is always normalized to lowercase keys when needed on Windows.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

compnerd (changes requested):
I think that the version that I put up is better.  It has the benefit of *preserving* the case while giving the case insensivity.  That is the behaviour that we want, we do not want to change the case, we want case preserving, case-insensitive comparisons.

PR #454, opened 2023-12-18
TSCBasic: remove some deprecated interfaces in `Process`

compnerd:
These interfaces were deprecated in 2022, remove these now.  While there were some instances of the `init` still in use.  Thanks to @MaxDesiatov and @weissi for the help to verify the safe removal of these deprecated interfaces.

### DIFF
--- a/Sources/TSCBasic/Process/Process.swift
+++ b/Sources/TSCBasic/Process/Process.swift
@@ -227,6 +227,7 @@ public final class Process {
     private static let loggingHandlerLock = NSLock()
 
     /// Global logging handler. Use with care! preferably use instance level instead of setting one globally.
+    @available(*, deprecated, message: "use instance level `loggingHandler` passed via `init` instead of setting one globally.")
     public static var loggingHandler: LoggingHandler? {
         get {
             Self.loggingHandlerLock.withLock {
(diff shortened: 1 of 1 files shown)


Sources/TSCBasic/Process/Process.swift
```diff
@@ -239,45 +239,7 @@ public final class Process {
...
-    @available(*, deprecated)
-    public var verbose: Bool {
-        get {
-            self.loggingHandler != nil
-        }
-        set {
-            self.loggingHandler = newValue ? Self.logToStdout : .none
-        }
-    }
+    public let loggingHandler: LoggingHandler?
```
MaxDesiatov, line 242: Without `private(set)` this will fail to build due to the `static` version of `loggingHandler`. Let's deprecate the static one while we're at it.

```suggestion
    public private(set) let loggingHandler: LoggingHandler?
```

compnerd: Odd, it seems to be building just fine for me locally.

compnerd:

Sources/TSCBasic/Process/Process.swift
MaxDesiatov: Ah, you're right, the static variable uses a separate static underscored property...

neonichu (approved):

compnerd:
@swift-ci please test

MaxDesiatov (approved):
Thank you 👍

MaxDesiatov:
@swift-ci test windows

compnerd:

PR #463, opened 2024-02-03
Update terminal width when progress animation is updated

kateinoigakukun:
Without updating the terminal width for each progress update, a progress line may be printed as more than one line, so line clearing will not work as expected and many lines will be printed when overflowing max columns of line.

Ninja also checks the terminal width every line print: https://github.com/ninja-build/ninja/blob/fd7067652cae480190bf13b2ee5475efdf09ac7d/src/line_printer.cc#L110

This particularly improves the build and test progress report of SwiftPM.

**Before**

https://github.com/apple/swift-tools-support-core/assets/11702759/7c32c2d5-0469-479b-a542-6ffd0656610a

**After**

https://github.com/apple/swift-tools-support-core/assets/11702759/71e6ce59-f888-4a19-819a-3012c9ea378f

### DIFF
--- a/Sources/TSCBasic/TerminalController.swift
+++ b/Sources/TSCBasic/TerminalController.swift
@@ -64,7 +64,14 @@ public final class TerminalController {
     private var stream: WritableByteStream
 
     /// Width of the terminal.
-    public let width: Int
+    public var width: Int {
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci Please test

PR #464, opened 2024-02-03
Put cursor at the end of last line for lit progress animation

kateinoigakukun:
The cursor was placed at the middle of the first progress line, and it makes the progress bar line too flickering (it's just my feeling). Also the last `Testing ` line was not printed when `Ctrl+C` signal is sent.

https://github.com/apple/swift-tools-support-core/assets/11702759/5c371ed4-1564-4a24-9ee4-59692fe99b30


kateinoigakukun:
@swift-ci Please test

MaxDesiatov (approved):
Good catch, thanks! I wonder if there's a good way to add a test case for it?

kateinoigakukun:
We need a complete terminal emulator to ensure the rendered cursor position, so there is no simple way, unfortunately.

MaxDesiatov:
That's fair, ready for merging then.

PR #472, opened 2024-04-24
Fix usage of Windows path API.

jeffdav:
We must call the API with `nil` first to get the size of the buffer, then allocate that amount. 

This fixes a crash in `swift-driver` on Windows.


compnerd (approved):
The change itself seems fine, though I'm not too big a fan of the whitespace changes.

compnerd:
@swift-ci please test

PR #479, opened 2024-08-02
[windows] path: drop the trailing backslash before invoking PathCchRe…

hyp:
…moveFileSpec to compute the dirname

PathCchRemoveFileSpec's is documented to not remove the file spec if it comes with a backslash suffix. Thus, we should remove the trailing backslashes before calling this API, to ensure we actually drop the current file/directory name.

This regressed after adoption of new swift-foundation, as it's filesystemRepresentation no longer drops the trailing backslash.


hyp:
@swift-ci please test

Sources/TSCBasic/Path.swift
```diff
@@ -461,7 +461,21 @@ private struct WindowsPath: Path, Sendable {
...
+        // PathCchRemoveFileSpec.
+        var substring = path[path.startIndex..<path.endIndex]
+        while !substring.isEmpty && substring.utf8.last == UInt8(ascii: "\\") {
+            substring = substring.dropLast()
+        }
+        if !substring.isEmpty && substring.last != ":" {
+            // Drop the trailing '\', unless the string path only
+            // has '\', and unless the slashes are right after the drive letter.
+            path = String(substring)
+        }
```
compnerd, line 477: Doesn't this get something like `D:\\\` wrong? That should normalize to `D:\` I think.

hyp: Yeah for sure. I think  `PathCchRemoveFileSpec` could normalize it, but let me check. I'll add a test for it too.

hyp: I checked, it looks like PathCchRemoveFileSpec normalizes it, so we don't need to take care of it here.

hyp:
@swift-ci please tset

hyp:
@swift-ci please test

compnerd (approved):

xedin:

xedin (approved):
Thank you! @rintaro and I spent some time today figure out what is going on with Windows build of swiftpm to find this exact problem.

xedin:

hyp:
Thanks, I will cherry-pick this into swift 6.

hyp:
swift 6 PR: https://github.com/swiftlang/swift-tools-support-core/pull/480

PR #485, opened 2024-10-10
Strip \\?\ prefix from path generated by resolveSymlink

ahoppen:
This aligns the implementation of resolveSymlinks with that of Foundation.String._resolvingSymlinksInPath, which received this canonicalization fix in https://github.com/swiftlang/swift-foundation/pull/639.

### DIFF
--- a/Sources/TSCBasic/PathShims.swift
+++ b/Sources/TSCBasic/PathShims.swift
@@ -24,21 +24,27 @@ import Foundation
 public func resolveSymlinks(_ path: AbsolutePath) throws -> AbsolutePath {
 #if os(Windows)
     let handle: HANDLE = path.pathString.withCString(encodedAs: UTF16.self) {
-        CreateFileW($0, GENERIC_READ, DWORD(FILE_SHARE_READ), nil,
+        CreateFileW($0, GENERIC_READ, DWORD(FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE), nil,
                     DWORD(OPEN_EXISTING), DWORD(FILE_FLAG_BACKUP_SEMANTICS), nil)
     }
     if handle == INVALID_HANDLE_VALUE { return path }
     defer { CloseHandle(handle) }
 
     let dwLength: DWORD =
-        GetFinalPathNameByHandleW(handle, nil, 0, DWORD(FILE_NAME_NORMALIZED))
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/TSCBasic/PathShims.swift
```diff
@@ -24,20 +24,21 @@ import Foundation
...
+        GetFinalPathNameByHandleW(handle, nil, 0, DWORD(VOLUME_NAME_DOS))
     return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
         guard GetFinalPathNameByHandleW(handle, $0.baseAddress!, DWORD($0.count),
-                                        DWORD(FILE_NAME_NORMALIZED)) == dwLength - 1 else {
+                                        DWORD(VOLUME_NAME_DOS)) == dwLength - 1 else {
             throw FileSystemError(.unknownOSError, path)
         }
-        let path = String(decodingCString: $0.baseAddress!, as: UTF16.self)
+        // When using `VOLUME_NAME_DOS`, the returned path uses `\\?\`.
+        let path = String(decodingCString: $0.baseAddress!.advanced(by: 4), as: UTF16.self)
```
compnerd, line 41: Do we want an assertion that the prefix is present?

compnerd (approved):

Sources/TSCBasic/PathShims.swift
ahoppen: I was following what you did in https://github.com/swiftlang/swift-foundation/pull/639. I guess we should also add the check there.

ahoppen:
@swift-ci Please test

Sources/TSCBasic/PathShims.swift
ahoppen: Making the corresponding change to swift-foundation here: https://github.com/swiftlang/swift-foundation/pull/974

ahoppen:
@swift-ci Please test Windows

ahoppen:
@swift-ci Please test Windows

ahoppen:
macOS CI failure is unrelated. This change only affects Windows. macOS CI is failing because this is the only repo that build SourceKit-LSP using a 5.9 host compiler, but SourceKit-LSP requires 5.10 to build successfully. We didn’t hit this before because this repo hasn’t seen any changes in the last 2 months, which is when SourceKit-LSP started requiring Swift 5.10.

PR #487, opened 2024-10-21
Add Swift Testing helper functions

bkhouri:
In the process of converting XCTests to Swift Testing, add a few Swift Testing helper functions equivalent to the XCTest equivalents.

Unfortunately, I'm unable to write automated tests that verify the new helpers behave as expected as the Swift Testing framework does not currently allow writing tests that verify an issue was recorded. This is tracked under https://github.com/swiftlang/swift-testing/issues/155

### DIFF
--- a/Sources/TSCTestSupport/SwiftTestingHelpers.swift
+++ b/Sources/TSCTestSupport/SwiftTestingHelpers.swift
@@ -1,0 +1,48 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2024 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Testing
+
+import TSCBasic
+
+public func expectFileExists(
+    _ path: AbsolutePath,
+    sourceLocation: SourceLocation = #_sourceLocation
+) {
+    print("localFileSystem.isFile(path) ==> \(localFileSystem.isFile(path))")
+    #expect(
+        localFileSystem.isFile(path),
+        "Expected file doesn't exist: \(path)",
+        sourceLocation: sourceLocation
+    )
+}
+public func expectDirectoryExists(
+    _ path: AbsolutePath,
+    sourceLocation: SourceLocation = #_sourceLocation
+) {
+    #expect(
+        localFileSystem.isDirectory(path),
+        "Expected directory doesn't exist: \(path)",
+        sourceLocation: sourceLocation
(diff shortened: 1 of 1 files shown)


Sources/TSCTestSupport/SwiftTestingHelpers.swift
```diff
@@ -0,0 +1,27 @@
...
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2024 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Testing
```
MaxDesiatov, line 11: This will fail on CI, which uses Swift 5.8 and doesn't have `swift-testing` included.

MaxDesiatov (changes requested):

Sources/TSCTestSupport/SwiftTestingHelpers.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import TSCBasic
+
+public func assertFileExists(_ path: AbsolutePath) {
```
grynspan, line 15: I would recommend plumbing down a `sourceLocation` argument here so that the diagnostics are attributed to the test, not the helper function.

bkhouri: This is completed.

Sources/TSCTestSupport/SwiftTestingHelpers.swift
```diff
@@ -0,0 +1,48 @@
...
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Testing
+
+import TSCBasic
+
+public func assertFileExists(
```
grynspan, line 15: I mean, I'd spell it `expectFileExists()` :P

bkhouri: That makes sense. I'll update all function names.

Sources/TSCTestSupport/SwiftTestingHelpers.swift
bkhouri: Are the pipelines for the project using Swift 6.0 yet?

bkhouri:
Ready for review, though it may be blocked as not all pipeline builds run with Swift 6.0!

MaxDesiatov:
@swift-ci test

bkhouri:
Opted to no longer proceed with the PR.  The associated helper functions will be added to the repository where they are needed, in this case swift-package-manager

PR #489, opened 2024-10-23
Remove referenced projects from multiroot data file

ahoppen:
We only build and test swift-tools-support-core with this multiroot-data-file, so there’s no reason why it should include any other projects.

This should fix the CI failures on macOS.


ahoppen:
@swift-ci Please test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov (approved):

PR #491, opened 2024-10-28
Fix FILEPointer typealias for Musl

marcprux:
Currently, `FILEPointer` is typealiased to `OpaquePointer` for Musl:

```swift
#if os(Android) || canImport(Musl)
public typealias FILEPointer = OpaquePointer
#else
public typealias FILEPointer = UnsafeMutablePointer<FILE>
#endif
```

But building an executable that links against TSCBasic fails:

```
$ ~/Library/Developer/Toolchains/swift-6.0.1-RELEASE.xctoolchain/usr/bin/swift build --swift-sdk x86_64-swift-linux-musl

.build/checkouts/swift-tools-support-core/Sources/TSCBasic/Process/Process.swift:706:70: error: cannot convert value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_IO_FILE>') to expected argument type 'FILEPointer' (aka 'OpaquePointer')
 704 |             throw Process.Error.stdinUnavailable
 705 |         }
 706 |         let stdinStream = try LocalFileOutputByteStream(filePointer: fp, closeOnDeinit: true)
     |                                                                      `- error: cannot convert value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_IO_FILE>') to expected argument type 'FILEPointer' (aka 'OpaquePointer')
```

But removing the `canImport(Musl)` check allows it to compile without issue, so I assume it was put there for an earlier version of the Musl SDK (or that it was never tested).

### DIFF
--- a/Sources/TSCBasic/WritableByteStream.swift
+++ b/Sources/TSCBasic/WritableByteStream.swift
@@ -66,7 +66,7 @@ public extension WritableByteStream {
 // Public alias to the old name to not introduce API compatibility.
 public typealias OutputByteStream = WritableByteStream
 
-#if os(Android) || canImport(Musl)
+#if os(Android)
 public typealias FILEPointer = OpaquePointer
 #else
 public typealias FILEPointer = UnsafeMutablePointer<FILE>


MaxDesiatov:
@swift-ci test

al45tair (approved):

al45tair:
> But removing the canImport(Musl) check allows it to compile without issue, so I assume it was put there for an earlier version of the Musl SDK (or that it was never tested).

Almost certainly it was added as part of @MaxDesiatov's earlier efforts to make Swift work with Musl, which did indeed use an older version of Musl but which also didn't fix up the headers that way the Static SDK for Linux does.

We're aware that we need more test coverage for the Static SDK as well, which would help us to catch things like this; there are plans to get some CI jobs up and running to that end (but maybe not PR testing just yet).

PR #493, opened 2024-11-19
[cmake] Install in a platform-specific directory

Steelskin:
The current shared library installation setup only installs the shared library into `/lib`. Since this is a Swift library, it should be installed in `/lib/swift/${os}` along with the swiftmodule directory.


etcwilde:
> Since this is a Swift library, it should be installed in `/lib/swift/${os}` along with the swiftmodule directory.

This is actually not really a great practice that Swift has fallen into (you wouldn't see <insert random C++ library name> inside of `/usr/lib/c++`. `<prefix>/lib/swift` should really be reserved for the swift standard library, though that ship has kind of sailed. The `${os}` is also an oddity since an SDK should really describe a single platform (if we see a `macosx/` directory in an SDK describing Debian, we've messed up somewhere). I'd prefer that we push folks toward using the values that `GNUInstallDirs` sets, and then adjusting if necessary. It knows how to set the install directories for the various Linux distros, which certainly have opinions about where things should go.

@compnerd, I know you have opinions on these things too.

Steelskin:
I had an offline chat with @compnerd and we agreed that the dynamic libraries should be kept to the directory from `GNUInstallDirs`, while the swiftmodules should still be kept to the per-target directory under `lib/swift/${os}`. I am closing this PR and will open a new one with the correct changes.

PR #502, opened 2025-03-28
Add a function on `LocalFileOutputByteStream` to disable `SIGPIPE` when writing data to it after the receiving end is closed

ahoppen:
Without this, if we launch a subprocess and try to write data to its stdin after it has finished (or crashed), we would take down the current process due to the SIGPIPE, which is usually not desirable. Eg. SourceKit-LSP doesn’t want to exit with a SIGPIPE exit code if a launched swift-format subprocesses crashes before the file to format could transferred to its stdin.

This mirrors behavior of `DispatchIO`, which also doesn’t emit a `SIGPIPE` when trying to write data to a stream whose receiving end has closed.

This will be used by https://github.com/swiftlang/sourcekit-lsp/pull/2098.

### DIFF
--- a/Sources/TSCBasic/WritableByteStream.swift
+++ b/Sources/TSCBasic/WritableByteStream.swift
@@ -817,6 +817,25 @@ public final class LocalFileOutputByteStream: FileOutputByteStream {
             throw error
         }
     }
+
+    #if canImport(Darwin)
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

MaxDesiatov (approved):

PR #513, opened 2025-06-16
build: use library directories instead of library names

compnerd:
`Foundation` was not being resolved to the target and instead would add `-lFoundation` to the linker invocation which is unnecessary due to the autolinking. Add the library search directory instead.


compnerd:

jakepetroules (approved):

PR #514, opened 2025-06-16
build: add TSCSupport to the built module set

compnerd:
Add TSCSupport to the CMake build system. This is meant to support the testing of SPM and not for general usage. We do not support the installation of the `TSCSupport` module as the usage is limited for the test suite in SPM.


jakepetroules (approved):

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

compnerd:
Ah, it seems that the XCTest framework is not being searched on macOS. This will require a small tweak.

PR #522, opened 2025-09-24
OpenBSD 7.8 will require OpaquePointer for FILE.

3405691582:
We changed Swift to use OpaquePointer in Glibc since OpenBSD 7.8 snapshots now hide the type information for FILE. Therefore, the types for the standard stdio streams should (regrettably) be OpaquePointer, due to the well-discussed issue of losing type information for forward-declared C types.

Subsequently, this means we should be using OpaquePointer here also.


3405691582:
See swiftlang/swift#84478 for the accompanying change.

PR #525, opened 2025-11-11
Remove redundant Sendable conformance

slavapestov:
ProcessEnvironmentBlock is a typealias to Dictionary, which is already Sendable when the key and value is Sendable.


slavapestov:
@swift-ci Please test

owenv (approved):

2tn6y27v66-svg (commented):
@

PR #534, opened 2025-12-09
Fix some cases where AbsolutePath::relative(to: AbsolutePath) would assert on Windows

daveinglis:
- there where a couple case where relative(to:) would assert on windows when a path was long (>260) and when tailing slashes where not removed in some cases.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -88,7 +88,7 @@ public struct AbsolutePath: Hashable, Sendable {
             }
             defer { LocalFree(pwszResult) }
 
-            self.init(String(decodingCString: pwszResult, as: UTF16.self))
+            try self.init(validating: String(decodingCString: pwszResult, as: UTF16.self))
 #else
             try self.init(basePath, RelativePath(validating: str))
 #endif
@@ -236,7 +236,7 @@ public struct RelativePath: Hashable, Sendable {
     fileprivate let _impl: PathImpl
 
     /// Private initializer when the backing storage is known.
-    private init(_ impl: PathImpl) {
+    fileprivate init(_ impl: PathImpl) {
         _impl = impl
     }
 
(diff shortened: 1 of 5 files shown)


daveinglis:
@swift-ci test

Sources/TSCBasic/Path.swift
```diff
@@ -515,13 +515,25 @@ private struct WindowsPath: Path, Sendable {
...
         }
+        // There seems to be many assumptions around paths and trailing '\'  
+        var substring = path[path.startIndex..<path.endIndex]
+        while !substring.isEmpty && substring.utf8.last == UInt8(ascii: "\\") {
+            substring = substring.dropLast()
+        }
+        if !substring.isEmpty && substring.last != ":" {
+            // Drop the trailing '\', unless the string path only
+            // has '\', and unless the slashes are right after the drive letter.
+            path = String(substring)
```
jakepetroules: Just to confirm - `C:` and `C:\` have completely different meanings. Is this logic preserving the trailing slash in the case where the path is pointing at a drive root?

jakepetroules:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test

Sources/TSCBasic/Path.swift
```diff
@@ -515,12 +515,28 @@ private struct WindowsPath: Path, Sendable {
     }
 
     init(string: String) {
-        if string.first?.isASCII ?? false, string.first?.isLetter ?? false, string.first?.isLowercase ?? false,
+        let path: String
+        let hasDrive: Bool
+        if string.first?.isASCII ?? false, string.first?.isLetter ?? false,
```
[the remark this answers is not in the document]
compnerd: I think that it not supporting that is fine. Foundation does go to some lengths to ensure that long paths are transparent to the user (it converts on the way in and out). Users generally would not type the NT path (and you cannot copy it from explorer as that transacts in Win32 paths and cmd would rely on DOS paths).

Sources/TSCBasic/Path.swift
```diff
@@ -950,23 +979,60 @@ extension AbsolutePath {
...
+            // This won't reconstruct correctly, but it's the best we can do.
+            if differentDrives {
+                let pathString = pathComps.joined(separator: "\\")
+                do {
+                    result = try RelativePath(validating: pathString)
+                } catch {
+                    // If that fails (because it contains drive letter), strip the drive
+                    // and use the remaining components
+                    let compsWithoutDrive = Array(pathComps.dropFirst())
+                    let pathWithoutDrive = compsWithoutDrive.joined(separator: "\\")
```
compnerd: This should be `([""]+compsWithoutDrive).joined(separator: "\\")`. `\directory\file.txt` vs `directory\file.txt` would be different when you restore the drive. That is:

`C:\directory\file.txt` vs `C:directory\file.txt`. These are two distinct paths, the latter being a drive directory relative path. Each drive has an associated cwd that the path would be relative to. The former is the absolute path formed by the current drive and thus a drive-relative path.

daveinglis:
@swift-ci test linux

compnerd (approved):
I think that there is a small edge case that isn't being considered, but this seems good outside of that.

PR #557, opened 2026-04-23
Fix some UNC path handling

daveinglis:
- use win32 api to strip prefix and preserve a drive letter based paths when resolving symlinks
- fix components to handle UNC paths.

### DIFF
--- a/Sources/TSCBasic/Path.swift
+++ b/Sources/TSCBasic/Path.swift
@@ -507,7 +507,25 @@ private struct WindowsPath: Path, Sendable {
         let normalized: UnsafePointer<Int8> = string.fileSystemRepresentation
         defer { normalized.deallocate() }
 
-        return String(cString: normalized).components(separatedBy: "\\").filter { !$0.isEmpty }
+        var str = String(cString: normalized)
+        // Normalize extended-length prefixes before classifying.
+        // \\?\UNC\server\share\... is a UNC path; \\?\C:\... is a local NT path.
+        if str.hasPrefix(#"\\?\UNC\"#) {
+            str = #"\\"# + str.dropFirst(7)   // \\?\UNC\server → \\server
+        } else if str.hasPrefix(#"\\?\"#) {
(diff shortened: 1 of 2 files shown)


daveinglis:
@swift-ci test

owenv (approved):
I think this is right but it would be good to get a second opinion before merging

Sources/TSCBasic/Path.swift
```diff
@@ -507,7 +507,18 @@ private struct WindowsPath: Path, Sendable {
...
+        let str = String(cString: normalized)
+        // UNC paths (\\server\share\...) start with two backslashes.  Preserve
+        // that as a synthetic "UNC" first component so that paths on different
+        // UNC servers/shares are never treated as descendants of each other, and
+        // so that UNC paths are never confused with local absolute paths that
+        // happen to share the same directory names.  "UNC" is a plain string
+        // with no backslashes, so it survives uses like lock-file name generation
+        // that join components with a separator character.
+        if str.hasPrefix("\\\\") {
+            return ["UNC"] + str.dropFirst(2).components(separatedBy: "\\").filter { !$0.isEmpty }
```
compnerd, line 526: This seems incorrect. `\\?\C:\` is not a UNC path, but this will treat it as a UNC path. You need to scan the prefix for making the determination I believe. If the idea is to tag the paths that start with `\\`, then UNC is the wrong marker, it should be `NTOrUNC`.

Note: it has to be a complete scan as `\\?\UNC\` is a UNC path as is `\\server\share\`, but `\\?\C:\` is a NT path.

daveinglis:
@swift-ci test

Sources/TSCBasic/Path.swift
daveinglis: right, ok, I added some pre-checks for these before this one.

compnerd (approved):

daveinglis:
@swift-ci test windows

PR #560, opened 2026-05-01
Fix Foundation.Pipe handle inheritance leakage on Windows

roman-bcny:
## Problem

On Windows, `Foundation.Pipe` creates handles with `bInheritHandle: true`. When multiple `Process` instances launch children concurrently, every `CreateProcess` inherits all currently-inheritable handles in the parent — including every other sibling's pipe write end. As a result, a pipe's read-side EOF only fires once the **slowest** sibling exits, since every sibling holds a copy of every other's write end open.

This is observable via `process.waitUntilExit()` returning clustered at the slowest-sibling exit time even when the underlying child finished much earlier. For workloads with many short-lived concurrent children (e.g. swift-driver dispatching swift-frontend invocations under `-j N`), this defeats the parent's ability to recycle parallel-job slots and gates new launches on the slowest already-running child.

## Fix

Inside `#if os(Windows)`, replace `Foundation.Pipe` with a non-inheritable pipe we own (`CreatePipe` with `bInheritHandle: false`). Around `process.run()`, take a new global `processRunLock` and toggle the WRITE handle's `HANDLE_FLAG_INHERIT` only for the duration of the run, so only the intended child inherits it; concurrent `run()` calls block on the lock and their pipes' write handles remain non-inheritable. After `run()`, close the parent-side WRITE FileHandles so the child becomes the sole writer and the read-side EOF fires when this specific child exits.

Non-Windows paths are unchanged.

## Out of scope

- `stdinPipe` is still a `Foundation.Pipe`; its READ end leaks to siblings the same way. Doesn't affect stdout/stderr EOF (write-side and read-side leakage are independent), but should be fixed for consistency in a follow-up.
- Other inheritable handles outside `Process`'s control still leak across all `CreateProcess` calls. The fully Windows-correct fix would use `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` via `InitializeProcThreadAttributeList` to specify exactly which handles each child inherits — but that requires changing how `Foundation.Process.run()` calls `CreateProcess`.

### DIFF
--- a/Sources/TSCBasic/Process/Process.swift
+++ b/Sources/TSCBasic/Process/Process.swift
@@ -23,6 +23,19 @@ import Dispatch
 
 import _Concurrency
 
+#if os(Windows)
+// Global lock used to serialize CreateProcess calls across all concurrent
+// Process instances. Inside the lock we briefly mark our pipe WRITE handles
+// inheritable so the specific child we're launching inherits them. Without
+// this serialization, every concurrent CreateProcess in the parent process
+// would inherit every other sibling's pipe handles (because Foundation.Pipe
+// creates handles with bInheritHandle:true and CreateProcess inherits all
+// currently-inheritable handles). The result is that a pipe's read-side
+// EOF only fires once the SLOWEST sibling exits, since every sibling holds
+// a copy of every other's write end open.
+fileprivate let processRunLock = NSLock()
+#endif
+
 /// Process result data which is available after process termination.
 public struct ProcessResult: CustomStringConvertible, Sendable {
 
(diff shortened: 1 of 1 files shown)


roman-bcny:
oops meant to share this with my team not here - this is a confirmed proof-of-concept hack to get around the inheritance issue in swift-driver that has a clear effect on SWBCore builds on Windows, and probably in many other places

jakepetroules:
> but that requires changing how Foundation.Process.run() calls CreateProcess.

Yep, that's exactly where the fix should land and is how swiftlang/swift-subprocess does it. I wouldn't want to take this change for TSC.

PR #561, opened 2026-05-06
[windows] add PDB files to install target

charles-zablit:
This is part of https://github.com/swiftlang/swift/issues/88869.


jakepetroules:
Does @compnerd's comment https://github.com/swiftlang/swift-toolchain-sqlite/pull/41#pullrequestreview-4237639106 apply here too?

charles-zablit:
> Does @compnerd's comment [swiftlang/swift-toolchain-sqlite#41 (review)](https://github.com/swiftlang/swift-toolchain-sqlite/pull/41#pullrequestreview-4237639106) apply here too?

It does but we can't apply it here because CMake does not support `TARGET_PDB_FILE` for Swift projects.

compnerd:
I'm worried that this is just adding more debt - we should fix CMake first - there is no guarantee that the PDB path is what is computed here.

compnerd (changes requested):
Where does `CMAKE_Swift_LINKER_SUPPORTS_PDB` come from? That is in the CMake namespace, but I'm not aware of this variable.

compnerd (changes requested):
Ah its internal, but its always set to true, so this is unnecessary.

Sources/TSCBasic/CMakeLists.txt
```diff
@@ -74,4 +74,10 @@ install(TARGETS TSCBasic
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin)
 
+if(BUILD_SHARED_LIBS AND CMAKE_Swift_LINKER_SUPPORTS_PDB)
```
compnerd:
```suggestion
if(BUILD_SHARED_LIBS)
```

Sources/TSCUtility/CMakeLists.txt
```diff
@@ -67,4 +67,10 @@ install(TARGETS TSCUtility
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin)
 
+if(BUILD_SHARED_LIBS AND CMAKE_Swift_LINKER_SUPPORTS_PDB)
```
compnerd:
```suggestion
if(BUILD_SHARED_LIBS)
```

charles-zablit:
@swift-ci please test

compnerd (approved):

charles-zablit:
@swift-ci please test

charles-zablit:
We need the `CMAKE_Swift_LINKER_SUPPORTS_PDB` check otherwise we get CMake errors on Linux and macOS.

compnerd:
What's the error? The optional is meant to avoid the errors.

PR #564, opened 2026-05-30
Remove unnecessary type casting

compnerd:
The compiler now imports these as their canonical type, allowing us to directly use without the explicit type cast.

### DIFF
--- a/Sources/TSCBasic/CMakeLists.txt
+++ b/Sources/TSCBasic/CMakeLists.txt
@@ -42,6 +42,7 @@ add_library(TSCBasic
   TemporaryFile.swift
   TerminalController.swift
   Thread.swift
+  "WinSDK+Overlay.swift"
   misc.swift)
 
 target_compile_options(TSCBasic PUBLIC
--- a/Sources/TSCBasic/Lock.swift
+++ b/Sources/TSCBasic/Lock.swift
@@ -95,11 +95,11 @@ public final class FileLock {
             let h: HANDLE = lockFile.pathString.withCString(encodedAs: UTF16.self, {
                 CreateFileW(
                     $0,
-                    UInt32(GENERIC_READ) | UInt32(GENERIC_WRITE),
-                    UInt32(FILE_SHARE_READ) | UInt32(FILE_SHARE_WRITE),
+                    GENERIC_READ | GENERIC_WRITE,
+                    FILE_SHARE_READ | FILE_SHARE_WRITE,
                     nil,
-                    DWORD(OPEN_ALWAYS),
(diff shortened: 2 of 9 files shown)


compnerd:
@swift-ci please test

owenv:
Do we need to stage this in somehow? Unfortunately a lot of the dependents still have swift 6.2 CI jobs

compnerd:
@owenv no, this should be backwards compatible. I'm not sure why the Linux build is impacted by this, this should only be touching the Windows paths. The `WinSDK+Overlay.swift` is the backwards compatibility. Once we can move to 6.5, we should be able to drop that.

compnerd:
@swift-ci please test

compnerd:

owenv (approved):
ok, I think this should be fine then so long as the tests pass

owenv:
The linux failures are happening because we stopped producing toolchains for AL2 and the workflow YAML is out of date. UB9 is failing because the latest dev snapshot skipped it when publishing. Not caused by this change

jakepetroules (approved):

compnerd:
@swift-ci please test

compnerd:

compnerd:

compnerd:

compnerd:
```
23:46:57  /home/build-user/swift-argument-parser/Sources/ArgumentParser/Parsing/CommandParser.swift:305:11: error: consecutive declarations on a line must be separated by ';'
23:46:57    ) throws(CommandError) -> ParsableCommand {
23:46:57            ^
23:46:57            ;
```

Seems like an infrastructure issue on Linux @owenv @jakepetroules

owenv:
This job is using Swift 5.9 for some reason, I'm looking at why

compnerd:

carlpeto:

carlpeto:
I want to get fresh error messages on Windows...

compnerd:
@carlpeto this set passes fine, it just is waiting on the CI to correct the bootstrap version on a non-windows platform.

carlpeto:
> @carlpeto this set passes fine, it just is waiting on the CI to correct the bootstrap version on a non-windows platform.

I've checked out this branch but rebased on main and built locally on my PC. When I try to run Foundation tests with `build.ps1 .... -Test Foundation` I get a diagnostic saying something like...


```
let hDirectory: HANDLE = ".".withCString(encodedAs: UTF16.self) {
                            ^^ failed to produce a diagnostic
```

...isn't that the error we were talking about?

compnerd:
That is! Weird that I didn't see it in CI or locally. I wonder if there is some caching oddity.

PR #566, opened 2026-06-25
Fix readFileContents to support non-regular files via FileHandle fallback

daveinglis:
The Data API is used for regular files (mmap performance) and FileHandle is the fallback for pipes, FIFOs, and character devices that Data rejects.

cherrypicked from main so we don't regress in 6.4


daveinglis:
@swift-ci test

bkhouri (approved):

bkhouri:
@swift-ci test

